PR #2, opened 2024-01-20
[benson1029] iP

benson1029:
### Echon

> "Punctuality is the politeness of kings." - Louis XVIII of France [(source)](https://www.brainyquote.com/topics/punctuality-quotes)

Looking for a tool to remind you to do things on time? **Echon is what you need**! It's

- text-based
- ~~easy~~ VERY EASY to use

All you need to do is,

1. download it from the release page.
2. double-click it.
3. add your tasks.
4. let it manage tasks for you. 😄 

Features:
- [X] Adding tasks
- [X] Keeping track of deadline of tasks
- [ ] Export to other formats (coming soon)

---

```java
    ui = new UI();
    try {
        storage = new Storage(FILE_PATH);
        taskList = storage.load();
    } catch (DukeException e) {
        ui.printMessage("Error loading task list: " + e.getMessage());
        return;
    }
    commandCreator = new CommandCreator(taskList);
    ui.printWelcomeMessage();
    ui.processCommands(commandCreator, storage);
}
```


src/main/java/echon/Command.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    @Override
+    protected Task createTask() {
+        return new Todo(this.description);
+    }
+}
+
+class AddDeadlineCommand extends AddTaskCommand {
+    private String byDate;
```
chenyixin0: Maybe consider renaming this field to dueDate for greater clarity?

chenyixin0 (commented):
Overall, very well written with good adherence to the Java coding standard. Good incorporation of images in the chatbot as well!

src/main/java/echon/Command.java
benson1029: I was originally trying to follow the command format. But I agree that this variable name sounds unclear. I've renamed it to dueDate. Thanks for bringing that out!

src/main/java/echon/Command.java
```diff
@@ -0,0 +1,293 @@
+package echon;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+
+/**
+ * Represents a command that can be executed by the user.
+ */
+public abstract class Command {
```
gowthaman-01, line 9: Avoid declaring multiple Java classes in a single file.

src/main/java/echon/Task.java
```diff
@@ -0,0 +1,53 @@
...
+    private boolean isDone;
+
+    /**
+     * Creates a task with the given description.
+     *
+     * @param description Description of the task.
+     */
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
```
gowthaman-01, line 17: Try to reduce the use of this. Use of this is discouraged, potentially to avoid unnecessary noise in the code. Try to avoid using this except when it is shadowed by a method or constructor parameter.

src/main/java/echon/TaskList.java
```diff
@@ -0,0 +1,107 @@
...
+        if (index < 0 || index >= this.tasks.size()) {
+            throw new EchonException("OOPS!!! The task index is invalid.");
+        }
+        return this.tasks.get(index);
+    }
+
+    /**
+     * Returns the number of tasks in the task list.
+     *
+     * @return The number of tasks in the task list.
```
gowthaman-01, line 102: The `@return` doesn't add any meaning, so you can remove it.

src/main/java/echon/TaskBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+     * @throws EchonException If the line is invalid.
+     */
+    public static Task createTaskFromFileLine(String line) throws EchonException {
+        String[] lineSplit = line.split(" \\| ");
+        String type = lineSplit[0];
+        String status = lineSplit[1];
+        String description = lineSplit[2];
+        Task task;
+        switch (type) {
+        case "T":
```
gowthaman-01, line 34: Try using an enum instead of matching strings

PR #12, opened 2024-01-24
[cyqjoseph] iP

cyqjoseph:
# RoeBot

> I could not come up with a better name

RoeBot frees your mind of having to remember things you need to do. It's
- Easy to use
- Free to use
- Fast to use
All you need to do is,
1. Download it from [here](https://github.com/nus-cs2103-AY2324S2/ip)
2. Double-Click it
3. Add your tasks
4. Let it manage you tasks for you :+1:

And it is once again, **FREE**

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)

***
If you are a Java Programmer, you can use it to practice Java too. Here's the `code`
```
public class Main {
    }
}
```


src/main/java/commands/AddEventCommand.java
```diff
@@ -0,0 +1,35 @@
...
+     */
+    public AddEventCommand(String name, LocalDateTime start, LocalDateTime end) {
+        this.name = name;
+        this.start = start;
+        this.end = end;
+    }
+
+    @Override
+    public UserCommand execute(TaskList taskList, UI ui, Storage storage) {
+        Event event = new Event(this.name, this.start, this.end);
```
miffi, line 31: `this.name`, `this.start`, and `this.end` don't need the `this` as there are no conflicting names in the local scope. According to the standard, it's better to leave the `this` out in this case, so  I think removing the `this` here and in some other parts of the code I'm seeing will be cleaner.

src/main/java/commands/DeleteTaskCommand.java
```diff
@@ -0,0 +1,30 @@
...
+import exceptions.DukeException;
+import services.Storage;
+import services.TaskList;
+import services.UI;
+import tasks.Task;
+
+/**
+ * Deletes a task from the task list.
+ */
+
```
miffi, line 12: The empty line between class and Javadoc header should be removed.

src/main/java/tasks/Deadline.java
```diff
@@ -0,0 +1,33 @@
...
+
+    /**
+     * Deadline constructor
+     */
+    public Deadline(String name, LocalDateTime by) {
+        super(name);
+        this.by = by;
+    }
+
+    public String formatBy() {
```
miffi, line 23: This method needs a javadoc header. I can't really understand what the function does just by looking at the name.

src/main/java/commands/AbstractCommand.java
```diff
@@ -0,0 +1,12 @@
...
+
+import services.Storage;
+import services.TaskList;
+import services.UI;
+
+/**
+ * Represents an executable command as an abstract class.
+ */
+public abstract class AbstractCommand {
+    public abstract UserCommand execute(TaskList taskList, UI ui, Storage storage);
```
miffi, line 12: Non-private method needs a javadoc header.

src/main/java/roebot/RoeBot.java
```diff
@@ -0,0 +1,47 @@
...
+     */
+    public RoeBot() {
+        this.storage = new Storage();
+        this.taskList = new TaskList();
+        this.ui = new UI();
+        this.parser = new Parser();
+        try {
+            storage.loadTasks(this.taskList);
+        } catch (IOException e) {
+            System.out.println("FIX THIS ERROR");
```
miffi, line 33: This got a chuckle out of me haha

src/main/java/services/Storage.java
```diff
@@ -0,0 +1,69 @@
...
+     * @throws IOException If an error occurs during file reading.
+     */
+    public void loadTasks(TaskList taskList) throws IOException {
+        assert taskList != null : "TaskList provided to loadTasks cannot be null";
+        try {
+            File file = new File(FILE_PATH);
+            if (!file.exists()) {
+                file.getParentFile().mkdirs();
+                file.createNewFile();
+                assert file.exists() : "File should exist after creation";
```
miffi, line 32: Nice use of asserts for checking invariants. Completely skipped my mind that they existed. I should use them as well.

src/main/java/tasks/Task.java
```diff
@@ -0,0 +1,40 @@
...
+    private String name;
+    /**
+     * Task constructor.
+     */
+    public Task(String name) {
+        this.name = name;
+        this.isDone = false;
+    }
+
+    public String getName() {
```
miffi, line 17: Missing JavaDoc header

miffi (commented):
Hello :). I'm a peer reviewer for your iP this week. This is my first review (the coding standard one).

As far as I can see, you have followed the coding standard really closely, so I don't have much to comment. I have just pointed out some nits (and some general comments), but otherwise LGTM!

src/main/java/services/Storage.java
cyqjoseph: Thanks for the comments!

PR #16, opened 2024-01-25
[Bandov] iP

Bandov:
### **BobBot**

> The More Time You Spend Awake The More Time You Spend Asleep. - Jaden Smith

Bob cooks, cleans and keeps your quintessential information for you. It's

- text-based
- fast-paced
- ~~easy~~ **super easy** to use

To get bob in your life you need to

1. download him from [here](https://github.com/Bandov/ip) 
2. run the installer
3. tell him your tasks
4. free money 😉

All for the cost of **$0**!

_Things bob can do!_

- [ ] cook
- [ ] clean
- [x] Manage tasks
- [x] Manage deadlines
- [x] Manage events

If you are a programmer come join me! Here is the main code for you to run
```java
public class Main {
    }
}
```


src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    /**
+     * Returns the string representation of the deadline.
+     *
+     * @return The string representation of the deadline.
+     */
+    @Override
+    public String toString() {
+        return "[D]" + (isDone ? "[X] " : "[ ] ") + super.description + " (by: " + by + ")"
```
GiselleTan, line 33: I noticed that you're using the (isDone ? "[X] " : "[ ] ") in the toString() and toFileString() functions as well as in other functions in other classes. Perhaps create a function such as printCurrentStatus() to accomplish this?

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    /**
+     * Creates an Event object from a string representation formatted for file storage.
+     *
+     * @param str The string representation of the event from a file.
+     * @return An Event object, or null if the string format is invalid.
+     */
+    public static Event fromFileString(String str) {
+        String[] parts = str.split(" \\| ");
```
GiselleTan, line 59: Naming it as "parts" can be confusing. Perhaps naming it as eventDetails would make it clearer what the String being split is.

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,109 @@
...
+        if (!parts[0].equals("E")) {
```
GiselleTan, line 60: Putting parts[0] in a variable might make it clearer what it is referring to and more readable.

GiselleTan (commented):
Overall, your comments helped with understanding your code a lot and the detail of the comments must have taken a lot of time!

src/main/java/duke/Ui.java
1rbg, line 1: Some minor inconsistency with the naming here, the first method is `void printWithLines()` which is fine and we expect it to print something behind the abstraction. The second method is `void showLoadingError()`, which is also fine since it will serve as a UI abstraction that shows the error on the UI. However, the last method is `String showWelcome()`, which also starts with show, which would indicate serving as a UI abstraction that shows the welcome, however it just returns a String. Perhaps it could be renamed `getWelcomeString()`, or the implementation could directly use `printWithLines()`.

src/main/java/duke/Task.java
1rbg, line 1: I like how you have named `toFileString()` and `fromFileString()` here, it is clear what it is used for and how it can be used.

src/main/java/duke/Storage.java
1rbg, line 1: I like that your imports are in alphabetical order and not using wildcard imports, this is consistent and follows the coding standard!

1rbg (commented):
Overall very well done! In terms of coding standard, I think you have done very well. Keep it up!

PR #25, opened 2024-01-25
[miffi] iP

miffi:
# The Cat that Lives in your Walls :cat:
```
 |\ /|
=(O O)=
 /   \
```
Just hand your tasks to it, and it'll keep track of it in the wall. It'll respond by gestures or by handing you notes. ~~It doesn't bite.~~

## Advantages of the mascot
* Cute (a given)
* Bit of a weirdo (who lives in _walls_?)
* **Cute**

## Why
Having a cat in your vicinity immediately raises your charisma which I've made into this handy formula.
```py
def effectiveCharisma(actualCharisma: int, hasCat: int) -> int:
    if hasCat:
        return actualCharisma + 13
    else:
        return actualCharisma
```

## Features
* [x] Tasks
* [x] Deadlines
* [x] Events
* [ ] GUI
* [ ] Editing tasks

## Installation and running
1. Have a java JDK installed
1. Go to the [releases](https://github.com/miffi/ip/releases) tab
1. Get the `jar` file
1. Run the jar file (either by clicking on it or by running it from the command line with `java -jar duke.jar`)

P.S. The cat told me to also say, and I quote:
> Meow

P.S.S. The charisma joke is inspired from https://www.youtube.com/watch?v=Ny5f1BsQsPE


src/main/java/cat/Cat.java
```diff
@@ -0,0 +1,67 @@
...
+     */
+    public void save() {
+        try {
+            storage.writeTaskList(tasks);
+        } catch (FileNotFoundException e) {
+            System.out.println("Cannot find state file \"" + FILE_NAME + "\"");
+        } catch (IOException e) {
+            System.out.println("Cannot write to state file \"" + FILE_NAME + "\"");
+        }
+    }
```
myang2020, line 66: Great that you separated a save() function from main function. Could improve even more by putting save() in a storage class.

src/main/java/cat/Parser.java
```diff
@@ -0,0 +1,221 @@
...
+        Todo,
+        Deadline,
+        Event,
+        Mark,
+        Unmark,
+        Delete,
+        Find,
+        List,
+        Help,
+        Bye;
```
myang2020, line 32: Great use of enums to denote a valid set of commands!

src/main/java/cat/Parser.java
```diff
@@ -0,0 +1,221 @@
...
+     */
+    public static class InvalidCommandData extends Exception {
+        public InvalidCommandData() {
+            super("Parameters to command not given.");
+        }
+
+        public InvalidCommandData(String key) {
+            super("No value given to " + key);
+        }
+    }
```
myang2020, line 210: Perhaps create a custom DukeException class

myang2020 (commented):
Sorry, reviewed wrong person.

src/main/java/cat/Parser.java
```diff
@@ -0,0 +1,221 @@
...
+        case Unmark:
+            return new CompleteCommand(parseIndices(input), false);
+        case Delete:
+            return new DeleteCommand(parseIndices(input));
+        case Find:
+            return new FindCommand(input);
+        default:
+            throw new IllegalStateException("Unexpected type: " + type);
+        }
+    }
```
0nesheep, line 101: I really like how you handled the command inputs! It's so clean 👍

src/main/java/cat/Parser.java
0nesheep, line 204: Might it be better to extract these exceptions to their own separate classes? Or at least a CatException class?

src/main/java/cat/TaskList.java
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * An exception used to signal that a task from a given request doesn't exist.
+     */
+    public static class TaskNotFound extends Exception {
+        public TaskNotFound(int index, Throwable cause) {
+            super("Could not find task " + (index + 1), cause);
+        }
+    }
+}
```
0nesheep, line 94: Similar to other comment on exceptions.

src/main/java/cat/command/AddCommand.java
```diff
@@ -0,0 +1,68 @@
...
+        Todo,
+        Deadline,
+        Event,
+    }
+
+    /**
+     * Constructs a command that adds a task to the task list.
+     *
+     * @param type       the type of the task
+     * @param components the components parsed from the user command
```
0nesheep: I think we need to capitalize the first letter and end the statement with a ```.```? Similar comment applies to other Javadoc, will not repeat to avoid too much repetition!

0nesheep (commented):
LGTM! Your error handling is very well done! Just some minor nit on extracting exception classes and a minor Javadoc styling error.

Also the cat is so cute <3 but get it out of my wall please :(

src/main/java/cat/command/AddCommand.java
miffi: I don't know how I missed that guideline. Thanks for pointing that out. ^^

src/main/java/cat/Parser.java
miffi: Hmm I think that makes sense. I was making the exceptions near where they are thrown from, but it might be better to centralise them. Well I'm still not 100% sold on that, but I think having a CatException to indicate that this is for internal errors is something I'll absolutely implement. Thanks!

PR #34, opened 2024-01-26
[Lim Zheng Ting] iP

AL-ZT:
# Duke v0.1
> “I don't know what to say so I'm linking my github here” – Zheng Ting ([source](https://github.com/AL-ZT))

Duke frees your mind of having to remember things you need to do. It's,

- text-based
- Not so easy to learn
- SLOW SUPER SLOW to use

All you need to do is,

1. download it from [here](https://github.com/AL-ZT/ip/releases/tag/v0.1).
2. download the jar file and run it with `java -jar <file-directory>/Duke-0.1.jar`.
3. add your tasks.
4. question your existence😉

And it is **FREE!**

Features:

- [x] Managing tasks, deadlines, events
- [x] Find a task
- [ ] GUI (coming soon)

```Java
public class Main {
        new Duke("./data/duke.txt").run();
    }
}
```


src/main/java/duke/Command.java
```diff
@@ -0,0 +1,55 @@
...
+     * Constructs a valid <code>Command</code> for execution. It however, does not take into account
+     * the validity of arguments specified by the user.
+     *
+     * @param cmd Main command.
+     * @param args Supplied arguments.
+     */
+    protected Command(String cmd, String[] args) {
+        this.command = cmd;
+        this.arguments = args;
+    }
```
Pluiexo, line 22: Isn't it better to omit the this keyword because it is inherently understood where the command object belongs in this context?(This applies to other sections in the program as well)

src/main/java/duke/tasks/Task.java
```diff
@@ -0,0 +1,68 @@
+package duke.tasks;
+
+/**
+ * Bare task that has a task name and a boolean condition on whether the task is done.
+ */
+public class Task {
+    protected String taskName;
```
Pluiexo: Having task in task name seems abit redundant, wouldn't it be better to shorten it to name?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,158 @@
...
+        this.tasks = new ArrayList<>();
+    }
+
+    /**
+     * Constructs a <code>TaskList</code> with specified task list.
+     *
+     * @param tasks <code>ArrayList</code> of tasks.
+     */
+    protected TaskList(ArrayList<Task> tasks) {
+        this.tasks = tasks;
```
Pluiexo, line 34: This is a good example of when to keep the this keyword! From the previous comment message

Pluiexo (commented):
Overall, good formatting! Class names, variables names are up to standard and they convey a good message of what they represent. Although they may be some areas for improvement in some difficult areas of the formatting, its still good effort nonetheless!

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,101 @@
...
+                                )
+                        );
+                    } else if (t instanceof Event) {
+                        this.data.write(String.format("E|%s|%s|%s|%s\n",
+                                t.getTaskName(), t.isDone(), ((Event) t).getFrom(), ((Event) t).getTo()));
+                    }
+                } catch (IOException e) {
+                    System.out.println(e.getMessage());
+                }
+            });
```
Pluiexo: Wow this is really difficult to read, the level of indentation seems to be off and the brackets are not leveled properly! May I suggest reconfiguring your wrapping settings in your IDE auto formatter if you are using one. Also nits tl could be better phrased into its full name for better readability in this case to differentiate between t of task

Pluiexo (commented):
Change selected lines

src/main/java/duke/Command.java
AL-ZT: Will do, thanks! It's a force of habit of mine.

src/main/java/duke/Storage.java
AL-ZT: Thanks for your suggestion, I think I was merely following the checkstyle back then. If you don't mind me asking, where would the text wrap setting be in the IDE for this particular case? Much appreciated

Pluiexo: If you are using IntelliJ -> settings -> code style -> java -> wrapping and braces. You can tweak the settings there to your preference!

AL-ZT: Awesome, thanks a lot!

PR #35, opened 2024-01-26
[Tan Guan Quan] iP

guanquann:
**GuanGuanBot**

> "Effective task management does not mean doing more things, but doing the right things. It's not about managing time, but managing priorities." _- A wise man_

GuanGuanBot is a chatbot designed for efficient task management to streamline your daily tasks and enhance productivity. It's,
- text-based
- easy to learn
- ~FAST~ SUPER FAST to use

All you need to do is,

1. Download the `.jar` file from [here](https://github.com/guanquann/ip/releases/download/A-Release/guanguan.jar).
2. Copy the jar file into an empty folder.
3. Open a command window in that folder.
4. Run the command `java -jar guanguan.jar` (i.e., run the command in the same folder as the jar file).
5. Add your tasks.
6. Let it manage your tasks for you 😉
7. PROFIT🤑!

And it is **FREE**!

Features:

- [x] Managing tasks
- [x] Managing deadlines
- [x] Managing events
- [x] Auto save
- [ ] Reminders (coming soon! _hopefully..._)

If you are a Java programmer, you can also use it to practice Java. Here's the method using Gradle:

```java
public class Launcher {
        Application.launch(Main.class, args);
    }
}
```


src/main/java/guanguan/GuanGuan.java
```diff
@@ -0,0 +1,76 @@
...
+
+import java.util.Scanner;
+
+/**
+ * Main class of chatbot.
+ */
+public class GuanGuan {
+    private final Storage storage;
+    private final TaskList items;
+    private final Ui ui;
```
azumieflare, line 11: final variables should be in all caps

src/main/java/guanguan/Gui.java
```diff
@@ -0,0 +1,92 @@
...
+     * Get accumulated output from the UI.
+     * @return Text output accumulated so far.
+     */
+    public String getTextOutput() {
+        String out = this.buffer.toString();
+        this.buffer.delete(0, this.buffer.length());
+        return out;
+    }
+
+    @Override
```
azumieflare, line 28: Good note that javadocs is not required as it performs a similar job as the parent method that it is overriding.

src/main/java/guanguan/Gui.java
```diff
@@ -0,0 +1,92 @@
...
+     */
+    public Gui() {
+        this.buffer = new StringBuilder();
+    }
+
+    /**
+     * Get accumulated output from the UI.
+     * @return Text output accumulated so far.
+     */
+    public String getTextOutput() {
```
azumieflare, line 22: Maybe could add @override here? Or could remove ```getTextOutput()``` from ```Ui.java``` as it seems to do nothing

src/main/java/guanguan/Ui.java
```diff
@@ -0,0 +1,113 @@
...
+     * Prints divider line message
+     */
+    public void line() {
+        System.out.println("____________________________________________________________");
+    }
+
+    /**
+     * Prints empty line message
+     */
+    public void emptyLine() {
```
azumieflare, line 97: This seems like a redundant method? Could just do ```System.out.println();``` instead of making a method for this?

azumieflare (commented):
Overall looks good!

guanquann:
@azumieflare thank you for your comment!

PR #51, opened 2024-01-26
[Apzure] iP

Apzure:
### Duke2

> "Your mind is for having ideas, not holding them." – David Allen

DukePro frees your mind of having to remember things you need to do.  It is 

- text-based
- *easy* to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,

1. Download it from [here](https://github.com/Apzure/ip/releases/tag/A-Jar).
2. Double-click it.
3. Add your tasks.
4. Let it manage your tasks for you 😸 
5. And it is **FREE!**

## Features:
- [x] Managing tasks

`def function()`

```java
public class Main {
    }
}

### CODE REVIEW

File: src/main/java/Duke/InvalidDateFormat.java [hunk 1]
```diff
@@ -0,0 +1,7 @@
+package duke;
+
+public class InvalidDateFormat extends DukeException {
```
JerryO3: Should this exception name follow the existing naming pattern you've established i.e. [description]Exception?

```suggestion
public class InvalidDateFormatException extends DukeException {
```

src/main/java/Duke/IndexOutOfRange.java
```diff
@@ -0,0 +1,7 @@
+package duke;
+
+public class IndexOutOfRange extends DukeException {
```
JerryO3: Should this exception name follow the existing naming pattern you've established i.e. [description]Exception?

```suggestion
public class IndexOutOfRangeException extends DukeException {
```

src/main/java/Duke/Bird.java
```diff
@@ -0,0 +1,66 @@
+package duke;
```
JerryO3, line 1: I like how you've made your Duke equivalent class readable using strong OOP principles. Was pleasantly surprised to find that the main application driver is so concise!

src/main/java/Duke/CommandHandler.java
```diff
@@ -0,0 +1,158 @@
...
+     *
+     * @param userInput the command to execute
+     * @return the response of the chatbot
+     * @throws DukeException if the command is not recognized or if an error occurs while executing the command
+     */
+    public String executeCommand(String userInput) throws DukeException {
+        String[] words = userInput.split("\\s+");
+        Command command;
+        try {
+            command = Command.valueOf(words[0].toUpperCase());
```
JerryO3: Really like this implementation of command parsing! Noted that this allows commands to be case-insensitive 👍

src/main/java/Duke/Storage.java
```diff
@@ -0,0 +1,147 @@
...
+                task = new ToDo("ERROR");
+                System.out.println("Error: Decoding Error, task does not match any of the known types");           
+        } 
+        TaskList.addTaskSilent(task);
+        if (marked) {
+            TaskList.markTaskSilent(TaskList.listSize());
+        }
+    }
+}
+
```
JerryO3: Would it be better if you provided more documentation on how the store and decode works? I only realised that you were storing the task as a string directly in the toString() format of each Task and then using decode() to parse this stored string back into a representative ItemList object after cross referencing both your store() and decode() methods. 

There may be other ways of doing this which may warrant your documentation on how you've implemented storage functionality. Otherwise, its a pretty neat implementation of storing Tasks 👍

JerryO3 (commented):
LGTM, some nits include your naming conventions regarding exceptions. Otherwise your code largely adheres to the coding standards! Really liked your implementation of command parsing as well :)

src/main/java/Duke/InvalidDateFormat.java
Apzure: Thanks for all the comments! Your suggestions make sense, and I'll be sure to improve with those in mind!

PR #84, opened 2024-01-27
[chaaaaun] iP

chaaaaun:
# Plana

> “I will give you a million chances.” – Sensei

Plana frees your mind of having to remember things you need to do. It's,
- text-based
- _easy to learn_
- **SUPER FAST** to use

All you need to do is,

1. download it from [here](https://github.com/chaaaaun/ip/releases).
2. double-click it.
3. add your tasks.

And it is FREE!

Features:

- [ ] Managing tasks
- [ ] Reminders (coming soon)

If you are a `Java` programmer, you can use it to practice `Java` too. Here's the `main` method:

```java
        try {
            Plana p = new Plana();
            p.init();
        } catch (ClassNotFoundException e) {
            System.out.println("Invalid save file:\n" + e.getMessage());
        } catch (IOException e) {
            System.out.println("Storage failed with error:\n" + e.getMessage());
        }
    }
```


src/main/java/chatbot/Command.java
```diff
@@ -0,0 +1,171 @@
...
+    LIST("list"),
+    MARK("mark"),
+    UNMARK("unmark"),
+    DELETE("delete"),
+    TODO("todo"),
+    DEADLINE("deadline"),
+    EVENT("event"),
+    FIND("find");
+
+    private final String rep;
```
kaitinghh, line 25: Final fields should be all capsed?

src/main/java/chatbot/Command.java
```diff
@@ -0,0 +1,171 @@
...
+
+    /**
+     * Executes a command with arguments for the provided task list.
+     *
+     * @param view The UI object to interface with.
+     * @param tl The task list object to operate on.
+     * @throws DukeException If an error occurs with the execution.
+     */
+    private void executeAdd(Ui view, TaskList tl) throws DukeException {
+        Task t = new Task("");
```
kaitinghh: What is the purpose of this line?

kaitinghh (commented):
LGTM!

src/main/java/chatbot/Command.java
chaaaaun: You're right, thanks for spotting!

src/main/java/chatbot/Command.java
chaaaaun: To initialise a default task, in case none of the switch cases below are applicable, should not happen in normal situations.

src/main/java/chatbot/DeadlineTask.java
```diff
@@ -0,0 +1,28 @@
...
+
+import java.io.Serializable;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+import chatbot.exceptions.InvalidArgumentException;
+
+class DeadlineTask extends Task implements Serializable {
+    private static final long serialVersionUID = 1L;
```
ondretann, line 11: Will it be better if you caps it since its final? Otherwise, UID should not be in caps as well! But I do like how descriptive your constant is :)

src/main/java/chatbot/Plana.java
```diff
@@ -0,0 +1,43 @@
...
+import chatbot.exceptions.DukeException;
+
+/**
+ * Represents the main class responsible for running the chatbot 'Plana'. Critical components like UI, file IO are
+ * initialised here.
+ */
+public class Plana {
+    private final TaskList taskList;
+    private final Storage store;
+    private boolean shouldExit;
```
ondretann: Will it be better to name this isExit instead to follow the java coding standard?

ondretann (commented):
Apart from minor errors that deviate from the java coding standard, everything else LGTM!

PR #105, opened 2024-01-27
[casaarlai] iP

casaarlai:
# Lai
_A rather **cynical** personality that helps you keep track of your life._
> Write it down. Written goals have a way of transforming wishes into wants; cant's into cans; dreams into plans; and plans into reality. Don't just think it - ink it! – Michael Korda [(source)](https://www.azquotes.com/quote/856710)
  
Lai is currently able to keep track of
- [x] Events (with support for start and end dates)
- [x] Deadlines (with support for due dates)
- [x] To-do items
- [ ] Emotional ramblings (hopefully Lai never has to!)

Why use Lai?
1. Completely free :star2:
2. Extremely performant 
3. New features added frequently!

Some basic commands include:
+ ```list``` to retrieve your list of tasks saved
+ ```bye``` to exit Lai
+ ```mark n``` to mark the task at the _nth_ position in the list as done
+ ```unmark n``` to mark the task at the _nth_ position in the list as not done

## Lai was proudly implemented in Java!

``` java
/**
 * Constructs a new Event with the specified description, start date, and end date.
 * The start and end dates are parsed from strings. If parsing fails using ISO-8601 format, an attempt is made using
 * "dd MMM yyyy" format.
 *
 * @param description The description of the event.
 * @param from The start date of the event in ISO-8601 format or "dd MMM yyyy" format.
 * @param to The end date of the event in ISO-8601 format or "dd MMM yyyy" format.
 * @throws DateTimeParseException If the start or end date strings cannot be parsed.
 */
public Event(String description, String from, String to) {
    super(description);
    try {
        this.from = LocalDate.parse(from);
        this.to = LocalDate.parse(to);
    } catch (DateTimeParseException e) {
        this.from = LocalDate.parse(from, DateTimeFormatter.ofPattern("dd MMM yyyy"));
        this.to = LocalDate.parse(to, DateTimeFormatter.ofPattern("dd MMM yyyy"));
    }
}
```


src/main/java/Task.java
```diff
@@ -0,0 +1,31 @@
...
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public Task(String description, boolean isDone) {
+        this.description = description;
+        this.isDone = isDone;
+    }
+
+    public String getDone() {
```
Kappaccinoh: I'm wondering if `getDone()` might be a slightly confusing naming convention, I recognise that the code here is quite simple so I can immediately tell you are retreiving the `boolean isDone` field, but perhaps renaming the function to `getisDone()` might be clearer

Kappaccinoh (commented):
Well done overall - I have not qualms about your naming conventions after looking through all the files. Keep up the naming conventions as you expand and build up your Lai ChatBot!

src/main/java/Task.java
casaarlai: @Kappaccinoh is there a reason setter functions for boolean variables follow this convention:

void setFound(boolean isFound);

but getter functions do not?

src/main/java/lai/task/Deadline.java
```diff
@@ -0,0 +1,60 @@
...
+        }
+    }
+
+    /**
+     * Returns a string representation of the deadline, including its type, completion status, description, and due
+     * date.
+     *
+     * @return A string in the format "[D][completion_status] description (by: due_date)", where completion_status is
+     * "X" if the task is done, or a space character otherwise, and due_date is formatted as "dd MMM yyyy".
+     */
```
tamagochuuu, line 55: Perhaps it would be better to break some of the longer lines of comment into two lines so that they are not too long

src/main/java/lai/task/Event.java
```diff
@@ -0,0 +1,77 @@
...
+     * end date.
+     * The start and end dates are formatted as "dd MMM yyyy".
+     *
+     * @return A string in the format "[E][completion_status] description (from: start_date to: end_date)", where
+     * completion_status is "X" if the event is done, or a space character otherwise.
+     */
+    @Override
+    public String toString() {
+        DateTimeFormatter dateFormat = DateTimeFormatter.ofPattern("dd MMM yyyy");
+        return "[E]" + super.toString() + " (from: " + from.format(dateFormat) + " to: " + to.format(dateFormat) + ")";
```
tamagochuuu, line 75: Perhaps it would be best to separate this line into a two lines as it is quite long

src/main/java/Lai.java
```diff
@@ -0,0 +1,117 @@
...
+                    Parser.checkDescription(desc);
+
+                    // Separating the start from description
+                    String[] descFrom = desc.split(" /from ", 2);
+                    desc = descFrom[0];
+                    String from = "";
+                    if (descFrom.length > 1) {
+                        from = descFrom[1];
+                    }
+                    // Separating the end out
```
tamagochuuu: Perhaps this comment could be a little clearer

tamagochuuu (commented):
Good

src/main/java/lai/util/Storage.java
```diff
@@ -0,0 +1,112 @@
...
+                } else if (line.startsWith("[D]")) {
+                    int byIndex = description.indexOf(" (by: ");
+                    String byString = description.substring(byIndex + 6, description.length() - 1);
+
+                    tasks.add(new Deadline(description.substring(0, byIndex), isDone, byString));
+
+                } else if (line.startsWith("[E]")) {
+                    int fromIndex = description.indexOf(" (from: ");
+                    int toIndex = description.indexOf(" to: ");
+                    String fromString = description.substring(fromIndex + 8, toIndex);
```
E0735389, line 91: Using `split` might be better to avoid hard coding magic constants like 8 above.

src/main/java/lai/util/TaskList.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * Represents a list of tasks in the Lai application.
+ * This class provides functionality to add, remove, and access tasks,
+ * as well as to iterate over all tasks.
+ */
+public class TaskList implements Iterable<Task> {
+    protected List<Task> tasks = new ArrayList<>();
```
E0735389, line 15: It's not clear why the default value is needed since it's overwritten in every constructor.

src/main/java/lai/util/TaskList.java
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * Searches for tasks that contain the specified query in their description.
+     * This method iterates over all tasks in the list and checks if the task's description contains the query string.
+     * If a match is found, the task is added to the result list.
+     *
+     * @param query The string to search for within task descriptions.
+     * @return A list of {@code Task} objects whose descriptions contain the specified query.
+     */
+    public List<Task> find(String query) {
+        List<Task> result  = new ArrayList<>();
```
E0735389, line 105: There should only be one space before the `=`.

src/main/java/lai/util/TaskList.java
```diff
@@ -0,0 +1,101 @@
...
+     * @param storage The storage utility to update the tasks file.
+     * @return The updated list of tasks.
+     */
+    public List<Task> add(Task newTask, Storage storage) {
+        this.tasks.add(newTask);
+        storage.updateTasksFile(this);
+
+        Ui.printTaskAdded(newTask, this);
+
+        return this.tasks;
```
E0735389: Be careful, the caller may accidentally modify the internal variable `tasks` if you do this.

src/main/java/lai/util/Ui.java
```diff
@@ -0,0 +1,142 @@
...
+    public static void printBye() {
+        printLine();
+        System.out.println("Goodbye, we shall meet again. Hopefully.");
+        printLine();
+    }
+
+    /**
+     * Prints a decorative line to separate sections of output for better readability.
+     */
+    public static void printLine() {
```
E0735389: Consider "printHorizontalLine" or "printDecorativeLine". Since println in fact stands for printline.

src/main/java/lai/util/TaskList.java
ChrisHo1341, line 105:
```suggestion
        List<Task> result = new ArrayList<>();
```

Might be good remove a space to follow the coding standards

src/main/java/lai/util/TaskList.java
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * Adds a new task to the list and updates the storage file.
+     * Also prints a message indicating the task has been added.
+     *
+     * @param newTask The task to be added.
+     * @param storage The storage utility to update the tasks file.
+     * @return The updated list of tasks.
+     */
+    public List<Task> add(Task newTask, Storage storage) {
```
ChrisHo1341: Consider renaming to `addTask() `to be even clearer

src/main/java/lai/util/TaskList.java
```diff
@@ -0,0 +1,101 @@
...
+        return this.tasks.size();
+    }
+
+    /**
+     * Retrieves a task by its position in the list.
+     *
+     * @param index The index of the task in the list.
+     * @return The task at the specified index.
+     */
+    public Task get(int index) {
```
ChrisHo1341, line 82: Consider renaming the function to `getTask()` to make it more clear

ChrisHo1341 (commented):
I found the code to be easy to read. Some function names are a bit ambiguous, consider changing them to make them clearer

PR #114, opened 2024-01-28
[kailinteoo] iP

kailinteoo:
### Kai

> Anyone who has never made a mistake has never tried anything new - Albery Einstein [source](https://dansilvestre.com/productivity-quotes)

**Introducing Kai: Your hassle-free task manager!**

- Text-based 
- Simple to master 
- Lightning-fast 

**Just:**
1. Download from [here](https://github.com/kailinteoo/ip/releases)
2. Double-click to launch 
3. Add your tasks 
4. Let Kassim handle the rest! 😎

**_And it's FREE!_** 

**Features:**

- [x] Add tasks, events, deadlines
- [ ] Reminders (coming soon)

***
if you are a Java programmer, you can use it to practice Java too. Here's the main method:

```java
public class Main {
    }
}
```


src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,75 @@
...
+                this.dayOfWeek = DayOfWeek.from(dayOfWeekFormatter.parse(by));
+
+            } catch (DateTimeParseException error) {
+                System.err.println("Error: Invalid date/time input. Please delete the task and provide a valid date/time or day of the week again.");
+            }
+        }
+    }
+
+    /**
+     * Returns a string representation of the Deadline task.
```
danielleloh: Since this is an overridden method, when writing Javadocs for it, you may want to consider using `@inheritDoc` tag to reuse the header comment from the parent method but with further modifications.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,82 @@
...
+        ArrayList<Task> myList = Storage.loadTasks();
+
+        if (myList == null) {
+            myList = new ArrayList<>();
+        }
+
+        // Initialize Scanner for user input
+        Scanner sc = new Scanner(System.in);
+        ui.blank();
+        boolean result = true;
```
danielleloh: Maybe you can consider renaming this boolean variable to sound like a boolean by using a name with a prefix such as `is`, `has`, `was`, etc.? Same for your other boolean method, parseCommand.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,201 @@
...
+            if (task.getDescription().contains(keyword)) {
+                System.out.println((myList.indexOf(task) + 1) + "." + task);
+            }
+        }
+
+        ui.separationLine();
+
+    }
+
+    public void add(String command, String restOfInputs) {
```
danielleloh: Could this method be named with more details such that it explains what this method is doing to readers?

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,187 @@
...
+        System.out.println(e.getMessage());
+        System.out.println("------------------------------------------------------------");
+    }
+
+    /**
+     * Returns a message indicating that a task description is empty.
+     *
+     * @return The error message for an empty task description.
+     */
+    public String emptyErrorMessage() {
```
danielleloh: Verbs should be used for method names so maybe you could change it to sound like a verb? (for e.g. `displayEmptyErrorMessage()`). Same for your other methods.

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,76 @@
...
+        try {
+
+            DateTimeFormatter dateTimeFormatter = DateTimeFormatter.ofPattern("d/M/yyyy HHmm");
+            this.by = LocalDateTime.parse(by, dateTimeFormatter);
+
+        } catch (DateTimeParseException e) {
+
+            try {
+                // If parsing as specific date/time format fails, try parsing as day of the week
+                DateTimeFormatter dayOfWeekFormatter = DateTimeFormatter.ofPattern("EEEE");
```
VatsalVitosh: Consider using constants for the date formats and declaring them as private static final fields at the top of the class. Doing so would make the method more generalised and if you needed to change the format, you would only need to change the constants instead of going to each method where it's used and changing it.

Example:
```ruby
private static final String DATE_TIME_FORMAT = "d/M/yyyy HHmm";
private static final String DAY_OF_WEEK_FORMAT = "EEEE";
```

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,87 @@
...
+    public boolean parseCommand() {
+
+        // Split input into "command" and "parameters"
+        String[] parts = userInput.split(" ", 2);
+        String command = parts[0];
+        String restOfInputs = parts.length > 1 ? parts[1] : "";
+
+
+        if (command.equals("list")) {
+            TaskList taskList = new TaskList(myList);
```
VatsalVitosh: Creating a new `TaskList` instance in each condition is unnecessary. Instead, perhaps create a single instance at the beginning of the method, and then use it as needed in each condition. This reduces the amount of duplicate code.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,201 @@
...
+
+    /**
+     * Marks a specified task in the list as done.
+     *
+     * @param parts An array containing the command and task number.
+     */
+    public void markList(String[] parts) {
+        //System.out.println("");
+        int index = Integer.parseInt(parts[1]) - 1;
+        Ui ui = new Ui();
```
VatsalVitosh: Similar to the `TaskList` instances, instead of creating a new `Ui` instance within each method, consider creating a single instance in the constructor and using it throughout the class.

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,189 @@
...
+        System.out.println("SORRY! but are you sure you enter the correct command? please check!");
+    }
+
+    /**
+     * Displays information that a task has been marked as done
+     *
+     * @param task The task that has been marked as done.
+     */
+    public void markInfo(Task task) {
+        System.out.println("------------------------------------------------------------");
```
VatsalVitosh: Consider using a constant for the separation line to ensure consistency across different methods.

Example:
```ruby
private static final String SEPARATION_LINE = "------------------------------------------------------------";
```

Doing this, you won't have to copy/paste the line over and over again and can just
```ruby
System.out.println(SEPARATION_LINE);
```

VatsalVitosh (commented):
Overall the code is very well constructed and closely follows the coding standard! I've only made a few suggestions to potentially make the code a little more concise :)

src/main/java/duke/Ui.java
kailinteoo: okay, I have changed it

src/main/java/duke/Deadline.java
kailinteoo: I will take it into consideration, thankyou

src/main/java/duke/Parser.java
kailinteoo: I have edited it accordingly. Thankyou

src/main/java/duke/TaskList.java
kailinteoo: done!

PR #142, opened 2024-01-28
[ibnu2651] iP

ibnu2651:
Ada frees your mind of having to remember things you need to do. It's,

- text based
- ~hard~ _easy_

All you need to do is,

1. Download it from [here](https://github.com/nus-cs2103-AY2324S2/ip/pull/142)
2. Run it 🏃 

Also **FREE**
Features:

- [x] Managing tasks
- [x] Not much else
---
Here's the `main` method:

```java
public class Ada {
        new Ada("ada.txt").run();
    }
}
```


src/main/java/adam/Parser.java
```diff
@@ -0,0 +1,101 @@
...
+     * Returns the command object after parsing through the user input.
+     *
+     * @param fullCommand The user input.
+     * @return The command object created.
+     * @throws AdaException If the user input is an invalid command or format.
+     */
+    public static Command parse(String fullCommand) throws AdaException {
+        String command = fullCommand.split(" ")[0];
+        String[] splitCommand;
+        String[] splitDesc;
```
angsongyee: Technically the variables here should be in plural form since they are of type `String[]` but I can't think of a suitable replacement name either😅.

src/main/java/adam/command/AddCommand.java
```diff
@@ -0,0 +1,69 @@
...
+import ada.ui.Ui;
+
+/**
+ * @inheritDoc
+ * Represents a command to add a task.
+ */
+public class AddCommand extends Command {
+    private String command;
+    private String desc;
+    private String[] details;
```
angsongyee, line 19: Good job on following the plural naming style here 👍

src/main/java/adam/task/Task.java
```diff
@@ -0,0 +1,65 @@
...
+     */
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // mark done task with X
+    }
+
+    /**
+     * Marks the task as done.
+     */
+    public void mark() {
+        this.isDone = true;
```
angsongyee: Its a good practice to omit the `this` keyword when accessing class members and you should only use it when there are ambiguous variable names in the scope (i.e. in constructors or when your method arguments have the same 
name)
```suggestion
        isDone = true;
```

angsongyee (commented):
Overall, the code is quite readable, and it follows the coding standard very well, other than a view minor nitpicks. Also, I noticed that currently there is some mismatch between the names in the project (e.g. filename says Adam.java but package and class name is Ada) but maybe it's because you haven't completely refactored everything yet so its not a problem for me. 👍 

(I just realized that I sent all my review comments individually instead of combining them, sorry for spamming your inbox 🙇)

ibnu2651:
Thanks for the helpful comments, yeah I'm not too sure what happened with the renaming, maybe doing it at 3 am was not a good idea. Don't worry about the multiple comments thing, it's all good. 👍

src/main/java/adam/Adam.java
shunjieee, line 1: Well abstracted and naming! No striking violation to the naming conventions. :)

PR #155, opened 2024-01-28
[Chin Zhe Ning] iP

chinzhening:
Mike frees your mind of having to keep track of things you need to do. It's,

- text-based
- easy to learn

All you need to do is,

1. download it from [here](https://github.com/biinnnggggg/ip/releases).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

***

```java
public class Main {
   }
}
```


src/main/java/mike/Mike.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    /**
+     * Starts the main control loop of Mike.
+     */
+    public void run() {
+        ui.displayWelcome();
+
+        boolean exitSeen = false;
+
```
vnnamng: Should this boolean flag be renamed?

src/main/java/mike/Storage.java
```diff
@@ -0,0 +1,118 @@
...
+        case "Event": // format: taskType,taskDescription,taskDone,startDate,endDate
+            String startDate = taskInformation[3];
+            String endDate = taskInformation[4];
+            newTask = new Event(taskDescription, startDate, endDate);
+            break;
+        default:
+            throw new MikeException("File corrupted. Unable to load saved data.");
+        }
+
+        if (taskIsCompleted) {
```
vnnamng: Name make sense but coding standard dont like...

src/main/java/mike/TaskList.java
```diff
@@ -0,0 +1,24 @@
...
+        return this
+                .stream()
+                .filter(task -> task.inListView(listView))
+                .map(task -> {
+                    int index = this.indexOf(task) + 1;
+                    return index + "." + task;
+                })
+                .reduce("", (s1, s2) -> s1 + "\n" + s2);
+    }
+
```
vnnamng: nice Stream bro

src/main/java/mike/Ui.java
```diff
@@ -0,0 +1,103 @@
...
+                "   #(     /##            *#,       (/\n" +
+                "   %*     /#(             #,       //\n" +
+                "   %(/.    %(            .%&      //*/\n" +
+                "   %/,&,   &(            #%/      (#//,\n" +
+                "   /(*     ##*           #%       &*#%\n" +
+                "     @/@*  (#*.          #(.      /%%\n" +
+                "      (/(/##(...........##(//&#(&/.\n" +
+                "     (@%(&@@,#,..........,*@@&&@&*,\n";
+        display(logo);
+    }
```
vnnamng: This logo should be saved as a static constant

src/main/java/mike/Ui.java
```diff
@@ -0,0 +1,103 @@
...
+        display(greeting);
+    }
+
+    /**
+     * Displays a horizontal line to the user.
+     */
+    public static void displayLine() {
+        String horizontalLine = "==========================================";
+        display(horizontalLine);
+    }
```
vnnamng: static constant same as the above, minor stuff no fetch

src/main/java/mike/CommandParser.java
```diff
@@ -0,0 +1,293 @@
...
+        return tokens.get(current - 1);
+    }
+
+    private boolean isAtEnd() {
+        return peekToken().getType() == TokenType.EOC;
+    }
+
+    private Token peekToken() {
+        return tokens.get(current);
+    }
```
vnnamng: the method names specific enough

vnnamng (commented):
good work bro!

src/main/java/mike/Storage.java
chinzhening: Nice catch, it will be changed.

src/main/java/mike/TaskList.java
chinzhening: Thanks! I am quite proud of it 😂😂

src/main/java/mike/CommandParser.java
chinzhening: My reasoning was that these methods are private and only meant for use inside the class so it would be specific enough in the context. What do you think?

PR #159, opened 2024-01-29
[Ng Wei Chong Javier] iP

javierng2knus:
> [!NOTE]
> This is still currently a work in progress project.

```
      ____.  _________   ____
     |    | /  _  \   \ /   /
     |    |/  /_\  \   Y   /
 /\__|    /    |    \     /
 \________\____|__  /\___/
                  \/
```

# Jav the Joy Amplifying Virtuoso 🎷

<details><summary>Some of the features</summary>
<p>

1. Greetings & Farewells
2. Stores tasks for you
   - To-dos (Do X)
   - Deadlines (Complete X by Y)
   - Events (Attend X from Y to Z)

</p>
</details> 

   
## How to use it
- [ ] Download the files
- [ ] ???
- [ ] **_Profit_**

Design patterns that are used:

- Singleton pattern using `LazyHolder` from [wikipage](https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom)

```java
public class Something {
    private Something() {}

    private static class LazyHolder {
        static final Something INSTANCE = new Something();
    }

    public static Something getInstance() {
        return LazyHolder.INSTANCE;
    }
}
```

> [!TIP]
> Be sure to let me know if there are any room for improvement! I'm always eager to learn more!

> The only true wisdom is in knowing you know nothing - Socrates [(source)](https://www.goodreads.com/author/show/275648.Socrates)

> [!WARNING]
> This is done on windows, mac and linux users _might_ face potential issues.

> [!CAUTION]
> My code is not perfect so refer to it at your own risk!


.vscode/settings.json
chiralcentre, line 1: Should the .vscode folder be removed from the repository? From my understanding, it is common practice to add the .vscode folder to the .gitignore file.

docs/README.md
chiralcentre, line 1: I like the level of detail in the README, which explains how to use the bot clearly

src/main/java/jav/Jav.java
chiralcentre, line 1: Should the Boolean variable isExiting be a local variable inside the main class instead? I was thinking it does not make sense for all instances of the Jav class to share the same value of isExiting.

src/main/java/jav/command/DeleteTaskCommand.java
chiralcentre, line 1: Would it be better to have different error messages to differentiate between the case where an integer was not provided and the case when an integer was provided but the integer is not in the valid range? Currently, for both cases, your code gives the same error message.

src/main/java/jav/command/UpdateTaskMarkCommand.java
chiralcentre, line 1: Would it be better to have different error messages to differentiate between the case where an integer was not provided and the case when an integer was provided but the integer is not in the valid range? Currently, for both cases, your code gives the same error message.

src/main/java/jav/manager/FileManager.java
chiralcentre, line 1: Would it be better to use StringBuilder in the load method? String concatenation is quite inefficient.

src/main/java/jav/task/Event.java
chiralcentre, line 1: Should the startDate and endDate attributes be changed to LocalDate objects, in accordance with Level 8 requirements? At the same time, should the toString() methods use String.format() instead of concatenation for greater clarity?

src/main/java/jav/task/Task.java
chiralcentre, line 1: Should Task be an abstract class at this point? From my understanding, there is no command to add a Task at the current stage, and thus there shouldn't be a need to instantiate an instance of the Task object. At the same time, the toString() methods for the subclasses of Task share a lot of similar code; would it be better to extract the common parts and place them in a helper method in Task instead?

chiralcentre (commented):
LGTM, but there are a few issues that I have highlighted, which I hope will be of use to you.

src/main/java/jav/Jav.java
```diff
@@ -0,0 +1,81 @@
...
+        // Load any neccessary data
+        Scanner scan = new Scanner(System.in);
+        try {
+            StorageManager.getInstance().load(FileManager.getInstance().loadStorageData());
+        } catch (Exception e) {
+            System.out.println(e.getMessage());
+        }
+
+        // Main program loop
+        while (!isExiting) {
```
jocelyn-soh: Nit: I think that using -ing form for boolean variables maybe a little contradictory since it's supposed to be only true or false but -ing refers to somewhere in between true and false. Perhaps you can consider variable names like isComplete, isFinished, etc.

src/main/java/jav/task/Event.java
```diff
@@ -0,0 +1,94 @@
...
+    protected String startText;
+
+    /** Non-date formatted end date. */
+    protected String endText;
+    
+    /**Date formatted start date. */
+    protected LocalDate startDate;
+
+    /**Date formatted end date. */
+    protected LocalDate endDate;
```
jocelyn-soh, line 23: I like the naming of these 4 variables! I think it's very clear and concise.

src/main/java/jav/command/UpdateTaskMarkCommand.java
```diff
@@ -0,0 +1,44 @@
...
+import jav.exception.InvalidParamException;
+import jav.manager.StorageManager;
+import jav.manager.UiManager;
+
+/**
+ * UpdateTaskMarkCommand handles the executing of updating tasks to marked/unmarked.
+ */
+public class UpdateTaskMarkCommand extends TaskCommand { 
+    /** Whether the task is going to be marked or unmarked. */
+    private boolean isMarking;
```
jocelyn-soh: Please refer to a similar comment about isExiting in Jav.java

jocelyn-soh (commented):
Code generally adheres to code quality guidelines. Other than a few nits, it LGTM! Good job!

PR #170, opened 2024-01-29
[irwynliong] iP

irwynliong:
Ever wanted a tasklist bot, that you could have used notion but instead **code it out yourself**? Introducing **Irwyn chatbot**!
It is :
- Not user friendly! 👍🏽
- ~~Effective~~ Not useful! 👍🏽
- Very Easy to use! 👎🏽

### All you need to do is 👇🏻

1. Clone the [Github repo](https://github.com/irwynliong/ip). 
2. Gradle run it. 🐘
3. List your task on the bot. 📄

And it just takes **_100000_** hours of your time to do this! ⌛ What a steal! 😎

### Upcoming updates:
- [x] Addition of GUI
- [x] Giving me headaches
- [x] Something

Heres a code snippet so that I meet the GFMD requirement, use `main` method:

```
        Irwyn.load();
        Irwyn.run();
    }
```


src/main/java/commands/Command.java
```diff
@@ -0,0 +1,41 @@
...
+import misc.StorageManager;
+
+/**
+ * This is an abstract class for all Commands used in the chatbot.
+ *
+ * @author  Irwyn Liong
+ * @version Week-3
+ */
+public abstract class Command {
+    boolean isExit;
```
chiageng: Good naming convention for a boolean value.

src/main/java/commands/CommandParser.java
```diff
@@ -0,0 +1,46 @@
...
+                return new MarkCommand(input);
+            case "unmark":
+                return new UnmarkCommand(input);
+            case "delete":
+                return new DeleteCommand(input);
+            case "find":
+                return new FindCommand(input);
+            default:
+                throw new TaskException();
+        }
```
chiageng: I noticed a slight mistake here, in CS2103T coding guides, switch keyword and case keyword should have same indentation.
```
switch (command) {
case "bye":
....
....
}

File: src/main/java/commands/ToDoCommand.java [hunk 3]
```diff
@@ -0,0 +1,42 @@
+package commands;
+
+import irwyn.tasks.TaskList;
+import irwyn.tasks.ToDo;
```
chiageng, line 4: It is good and clear that you listed out imported class explicitly.

src/main/java/irwyn/tasks/Task.java
```diff
@@ -0,0 +1,94 @@
...
+    protected String description;
+    protected boolean isDone;
+
+    TaskType task;
+
+    public enum TaskType {
+        TODO,
+        EVENT,
+        DEADLINE
+    }
```
chiageng, line 23: It is good that your constant variable name is all uppercase and is clear to other people.

chiageng (commented):
Overall it looks good. Keep it up.

LifHoshi (commented):
Well,  overall an excellent work of iP so far. Most of the obvious problems have been found out by another reviewer. By the way, I like the way that you create different folders to distinguish the use of purpose from other Java files.

PR #176, opened 2024-01-29
[danielleloh] iP

danielleloh:
# Scribbles
Scribbles helps you to track and manage the many tasks in your life! ✔️ 

It is:
- Text-based,
- Interactive,
- Simple and neat!
---
How to use Scribbles:
1. Click [here](https://github.com/danielleloh/ip) to download it
2. Launch it
3. And start using it!

The best thing about Scribbles is, it's **FREE** for all!

---
Features:
- [x] Manages your to-dos
- [x] Manages your deadlines
- [x] Manages your events
- [x] And more features coming up!

---
If you are learning Java, you can use it to practice your Java too! Here's the `main` method:
```java
public class Main {
    }
}
```
---
Remember,
> “Your mind is for having ideas, not holding them.” – David Allen


src/main/java/scribbles/Scribbles.java
```diff
@@ -0,0 +1,227 @@
...
+                String keyword = parsedInput.getFindKeyword();
+                ui.tasksWithKeyword(keyword, taskList);
+            } catch (IndexOutOfBoundsException e) {
+                ui.missingKeywordMessage();
+            }
+            break;
+        default:
+            ui.invalidInputMessage();
+        }
+        return false;
```
matthewken19: I think you can create your own exception and catch it at the end of the switch case statement so it will only handle your custom exception instead of handling all exceptions in every case.

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,124 @@
...
+                case "D":
+                    String deadlineString = tokens[3].trim();
+                    try {
+                        DateTimeFormatter dateTimeFormat = DateTimeFormatter.ofPattern("dd/MM/yyyy HHmm");
+                        LocalDateTime deadline = LocalDateTime.parse(deadlineString, dateTimeFormat);
+                        taskList.addTask(new Deadline(description, isCompleted, deadline));
+                    } catch (DateTimeParseException e) {
+                        System.out.println("Your deadline, \"" + description + "\" stored in file has incorrect "
+                                + "date/time format and cannot be loaded into task list. ");
+                    }
```
matthewken19, line 108: good handling of DateTimeParseException

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,124 @@
...
+                    writer.write("T | " + (task.isCompleted() ? "1" : "0") + " | " + task.getDescription());
+                }
+                if (task instanceof Deadline) {
+                    writer.write("D | " + (task.isCompleted() ? "1" : "0") + " | " + task.getDescription()
+                            + " | " + ((Deadline) task).getByString());
+                }
+                if (task instanceof Event) {
+                    writer.write("E | " + (task.isCompleted() ? "1" : "0") + " | " + task.getDescription()
+                            + " | " + ((Event) task).getStartString() + " | " + ((Event) task).getEndString());
+                }
```
matthewken19, line 158: I think you can include this method in the classes instead and use polymorphism from the Task class method

src/test/java/parser/ParserTest.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class ParserTest {
+
+    @Test
+    public void isMissingDeadlineInformationTest() {
```
matthewken19: I like how thorough you are with test cases! However I think naming convention should be something like: 
isMissingDeadlineInformation_correctFormat_success() or isMissingDeadlineInformation_incorrectFormat_exceptionThrown() 
to show unitBeingTested_descriptionOfTestInputs_expectedOutcome in the test name

matthewken19 (commented):
Overall, I like the OOP abstraction! LGTM, but just take note of some changes that I have proposed that might help!

src/main/java/task/Task.java
```diff
@@ -0,0 +1,54 @@
...
+        this.isCompleted = false;
+    }
+
+    /**
+     * Returns the String format of how tasks are listed in the list of task.
+     *
+     * @return User-readable task in String format
+     */
+    public String toString() {
+        if (this.isCompleted) {
```
brennalaurentan, line 57: Not a coding standard violation and this is fine, but can consider having a `completionIcon()` method which returns "[ ]" or "[X]" depending on the boolean isCompleted status instead of statically using if & else to determine which one to print

then this method can just be `return this.completionIcon() + " " + this.description`

will be neater if you decide to add more components to this `toString()` method in the future as you have more control over the individual components :)

src/test/java/scribbles/ScribblesTest.java
```diff
@@ -0,0 +1,49 @@
+package scribbles;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class ScribblesTest {
+
+    @Test
+    public void parseInputTest() {
```
brennalaurentan: Good use of JUnit to implement test cases! Comprehensive testing of software 👍

src/main/java/task/Deadline.java
```diff
@@ -0,0 +1,38 @@
+package task;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * This class represents a deadline
```
brennalaurentan: Maybe can provide a little more detail in the comment header for the class to indicate how Todo, Deadline and Event tasks differ? Such as mentioning how Deadline has a due date and Event has from and to dates

src/main/java/task/Event.java
```diff
@@ -0,0 +1,51 @@
+package task;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * This class represents an event.
```
brennalaurentan: Maybe can provide a little more detail in the comment header for the class to indicate how Todo, Deadline and Event tasks differ? Such as mentioning how Deadline has a due date and Event has from and to dates

brennalaurentan (commented):
Nice implementation of your own task manager! No violations found, but just a few suggestions here and there.

src/main/java/storage/Storage.java
danielleloh: Thank you :)

src/main/java/storage/Storage.java
danielleloh: Thank you for your advice :) I will try implementing that

src/test/java/parser/ParserTest.java
danielleloh: Thank you for the feedback:)

src/main/java/task/Task.java
danielleloh: Thank you for your advice! It really does seem neater if I implement it that way. I will try :)

src/test/java/scribbles/ScribblesTest.java
danielleloh: Thank you :)

src/main/java/task/Deadline.java
danielleloh: Thank you for your feedback! What you say makes sense and I will take it into consideration and edit it to be clearer :)

src/main/java/task/Event.java
danielleloh: Thank you for your feedback! What you say makes sense and I will take it into consideration and edit it to be clearer :)

PR #186, opened 2024-01-29
[Wang Xinrong] iP

wang-xinrong:
Benefits of using Duke include:
- text-based
- simplistic GUI
- **fun**

You can:
1. download it from [here](https://github.com/wang-xinrong/ip/releases/tag/A-Jar)

_It is **freeeee**_

Features:
- [x] Managing tasks
- [ ] Do tasks for you :satisfied:

it is developed using Java, take at look at this `main` function
`public class Main {
    }
}
`


src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,144 @@
...
+    private static Duke instance = null;
+    private LivState currentState = null;
+
+    /**
+     * You should have your own function to generate a response to user input.
+     * Replace this stub with your completed method.
+     */
+    public String getResponse(String input) {
+        try {
+            return parser.ProcessInputReturnString(input);
```
jeong-jaeho: Shouldn't the parser's method be in camelCase?

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,213 @@
...
+        return LocalDateTime.parse(input, DATE_TIME_FORMATTER);
+    }
+
+    /**
+     * Converts dataTime to duke command format.
+     *
+     * @param localDateTime DateTime to be converted.
+     * @return String representing dateTime in command format.
+     */
+    public static String convertDateTimeToCommandFormat(LocalDateTime localDateTime) {
```
jeong-jaeho, line 192: I like how you use name to explain what the variable is!

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,178 @@
...
+
+import duke.parser.MissingInputFieldException;
+
+/**
+ * Represents a task.
+ */
+public abstract class Task {
+    private boolean isDone;
+    protected TaskType type;
+    public static String delimiter;
```
jeong-jaeho: I like how you try to abide by the rule of using nouns for things and verbs for actions.

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,178 @@
...
+    public static String command;
+    protected static String dataStringSplitter = " \\| ";
+    // temporary measure before storage related methods are migrated over to the storage class
+    protected static String storageDataStringSplitter = " | ";
```
jeong-jaeho, line 11: I like how you try to avoid 'too short' names to explain what the variable does. It's not excessively lengthy, so it only helps me understand your code better.

src/main/java/duke/task/TaskList.java
```diff
@@ -0,0 +1,118 @@
...
+
+import duke.ui.Ui;
+
+/**
+ * Represents a taskList to store tasks.
+ */
+public class TaskList {
+    private static LinkedList<Task> tasks = null;
+    private static LinkedList<Task> lastFilteredTasks = null;
+    private static TaskList instance = null;
```
jeong-jaeho, line 16: Perhaps would it be better to name it more specific to the purpose it serves?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,144 @@
+package duke;
```
jeong-jaeho, line 1: Overall, I like how you use appropriate language and avoid slangs or names that are not meaningful within contexts/times.

jeong-jaeho (commented):
LGTM. Just some nits to fix.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,144 @@
...
+    /*
+    public static void main(String[] args) {
+        getInstance().instanceStart();
+    }
+     */
+
+    /**
+     * Toggles active state of Duke.
+     */
+    public void ToggleActiveState() {
```
ChillinRage: Is it better to use camelCasing for "toggleActiveState" to maintain naming consistency?

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,213 @@
...
+        if (words[0].equals("find")) {
+            taskList.findTaskWithKeyword(words[1]);
+            ui.listFilteredTasks();
+            return;
+        }
+
+        throw new CommandNotFoundException(input);
+    }
+
+    public String ProcessInputReturnString(String input) throws InputException {
```
ChillinRage: Is it better to use switch-case instead of if blocks?

src/main/java/duke/storage/Storage.java
```diff
@@ -0,0 +1,101 @@
...
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.time.LocalDateTime;
+import java.util.Scanner;
+
+import duke.parser.Parser;
+import duke.task.Task;
+import duke.task.TaskList;
```
ChillinRage, line 12: I like that the import packages are well separated and is in a sorted order.

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    public String printType() {
+        return "E";
+    }
+
+    @Override
+    public void setUpTask(String input) throws MissingInputFieldException {
+        try {
+            input = input.trim();
+            if (!input.contains(command)) throw new RuntimeException("not todo");
```
ChillinRage: Would it be better to follow the coding standard for an if block?

src/main/java/duke/ui/HorizontalLine.java
ChillinRage, line 1: I like that a horizontal line is also broken down into a class itself.

ChillinRage (commented):
Looks good!

PR #187, opened 2024-01-29
[Aw Joey] iP

AwJoey:
# Axolotl
>  “Your mind is for having ideas, not holding them.” – David Allen ([source](https://dansilvestre.com/productivity-quotes/))

Axolotl your mind of having to remember things you need to do. It's,
- text-based
* easy to learn
+ ~~FAST~~ *SUPER FAST* to use

All you need to do is,
1. download it from [here](https://github.com/AwJoey/ip).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:
- [X] Managing tasks
- [ ] Reminders (coming soon)

If you are a Java programmer, you can use it to practice Java too. Here's the ```main``` method:

```java
public class Main {
    }
}
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,193 @@
+import java.io.IOException;
+import java.lang.reflect.Array;
+import java.time.DateTimeException;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Scanner;
+import java.util.ArrayList;
+import  java.util.EnumSet;
+
+enum Command {
```
jinhanfromNUS: I like how you create this enum Command, it makes the code very readable.

src/main/java/Duke.java
```diff
@@ -1,10 +1,193 @@
...
+
+    public final String commandName;
+    public final String commandSignature;
+
+    private Command(String commandName, String commandSignature) {
+        this.commandName = commandName;
+        this.commandSignature = commandSignature;
+    }
+}
 public class Duke {
```
jinhanfromNUS: I feel like the part is a little bit too long. If there is another parser class, it would make the code much easier to understand.

src/main/java/DukeException.java
```diff
@@ -0,0 +1,3 @@
+public class DukeException {
```
jinhanfromNUS: I think it would be better if we delete this part, as it is empty and it might cause confusion in the future :).

src/main/java/Storage.java
```diff
@@ -0,0 +1,91 @@
...
+import java.nio.file.Paths;
+import java.time.DateTimeException;
+import java.time.LocalDateTime;
+import java.util.ArrayList;
+import java.util.List;
+
+public class Storage {
+
+    static String HOME = System.getProperty("user.home");
+    static Path FILEPATH = Paths.get(HOME, "Downloads", "NUS_CS",
```
jinhanfromNUS: I am not sure if I am missing something. This part looks like it is using an absolute path, which might easily create bugs for other users without the same path.

src/main/java/Storage.java
```diff
@@ -0,0 +1,91 @@
...
+            "CS2103", "duke.txt");
+
+    public static ArrayList<Task> retrieveList() {
```
jinhanfromNUS: Perhaps add a throw exception when file path is not found? 🤔

jinhanfromNUS (commented):
Keep up the good work!

src/main/java/Parser.java
```diff
@@ -0,0 +1,90 @@
+import java.time.DateTimeException;
```
chuahjiajie: Nice parser implementation, will use as inspiration.

chuahjiajie (commented):
Happy New Year

PR #196, opened 2024-01-29
[doyelee0313] iP

doyelee0313:


src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,108 @@
...
+    public TaskList(List<Task> tasks) {
+        this.tasks = tasks;
+    }
+
+    /**
+     * Constructs an empty TaskList.
+     */
+    public TaskList() {
+        this.tasks = new ArrayList<>();
+    }
```
Kaya3842: Why not use only ArrayList<Task>?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+     * Marks a task as done.
+     *
+     * @param number The number of the task to be marked as done.
+     */
+    public void mark(int number) {
+        Task t = tasks.get(number - 1);
+        t.markAsDone();
+    }
```
Kaya3842: Shouldn't these two methods belong to the Task class?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,145 @@
...
+     * Reads user commands and actions until the user enters "bye"
+     *
+     * @param args The command-line
+     */
+    private Storage storage;
+    private TaskList tasks;
+    private Ui ui;
+    private Parser parser;
+
+    private static String filePath = "/Users/leedoye/ip/src/data/duke_tasks.txt";
```
Kaya3842: I think the file path should not be absolute.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,54 @@
...
+import java.io.ObjectOutputStream;
+import java.util.List;
+
+/**
+ * Storage class load and save the task from a file.
+ * It provides methods to read and write the tasks
+ *
+ */
+public class Storage {
+    private String filePath = "/Users/leedoye/ip/src/data/duke_tasks.txt";
```
Kaya3842: Same issue regarding absolute filepath.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,54 @@
...
+     * @throws DukeException If there is an error loading tasks from storage.
+     */
+    public List<Task> load() throws DukeException{
+        try (ObjectInputStream intput = new ObjectInputStream(new FileInputStream("filePath"))) {
+            return (List<Task>) intput.readObject();
+        } catch (IOException | ClassNotFoundException e) {
+            System.out.println("No existing tasks or file.");
+            throw new DukeException("Error loading tasks from storage");
+        }
+    }
```
Kaya3842:
```suggestion
    public List<Task> load() throws DukeException{
        try {
            ObjectInputStream intput = new ObjectInputStream(new FileInputStream("filePath"));
            return (List<Task>) intput.readObject();
        } catch (IOException | ClassNotFoundException e) {
            System.out.println("No existing tasks or file.");
            throw new DukeException("Error loading tasks from storage");
        }
    }
```

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,54 @@
...
+     * @return The list of tasks saved from the file.
+     * @throws DukeException If there is an error saving tasks from storage.
+     */
+    public void save(List<Task> tasks) {
+        try (ObjectOutputStream output = new ObjectOutputStream(new FileOutputStream("filePath"))) {
+            output.writeObject(tasks);
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+    }
```
Kaya3842: Same issue with try-catch statements can be found more in this class.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,145 @@
...
+                TaskList newtasks = new TaskList(tasks.findTasks(word));
+                response = ui.findList(newtasks.getTasks());
+            } catch (DukeException e) {
+                response = e.getMessage();
+            }
+        } else {
+            response = ui.dontUnderstand();
+        }
+        return response;
+    }
```
Kaya3842: Should this be extracted out? For example you can extract one method for Parser class and one method for Ui class instead of combining them in the Duke class. Or create a new method only for extracting the numbers in a command.

Kaya3842 (commented):
Most of the classes was extracted out reasonably well. There was some minor coding standard violations but it can easily be fixed.  Consider extracting more methods in Parser and Ui class to make the main Duke class shorter.

PR #203, opened 2024-01-29
[Darryl Chan] iP

darryl-chan:
# LinusBot

> "Better responses than ChatGPT" - OpenAI CEO

Have you ever wanted a bot that is
-  Text-based
-  Keeps track of your tasks
-  ~~Easy~~ Very easy to use

Then you came to the right place 😉 

You just have to:
1. Download it [here](https://github.com/darryl-chan/ip)
2. run it
3. Its that easy

And might I remind you that its **FREE** to use

Current features:
- [x] Manage tasks
- [x] Cool graphics

And if you love java you can download this app and improve your skills. Here's the `Launcher` class:

```
public class Launcher {
        Application.launch(Main.class, args);
    }
}
```


docs/README.md
```diff
@@ -1,29 +1,30 @@
-# User Guide
+# Duke User Guide
```
Kappaccinoh, line 1: I see that you have actually renamed your ChatBot to Linus! It's a nice name! Might it be better to change your main class to Linus to avoid confusion with Duke?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,205 @@
+package duke;
```
Kappaccinoh: I see that you have actually renamed your ChatBot to Linus! It's a nice name! Might it be better to change your main class to Linus to avoid confusion with Duke?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,205 @@
...
+
+// import javafx.scene.control.Label;
```
Kappaccinoh: I understand that you have moved these comments out to another directory, I think you can delete these comments to keep the file cleaner and shorter

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,113 @@
...
+
+import duke.commands.CommandFind;
+import duke.tasks.TaskList;
+
+/**
+ * Class that reads user input
+ */
+public class Parser {
+
+    enum CommandType {LIST, BYE, MARK, UNMARK, DELETE, TODO, DEADLINE, EVENT, FIND};
```
Kappaccinoh, line 25: Nice use of Enums!

src/main/java/duke/commands/CommandBye.java
```diff
@@ -0,0 +1,23 @@
...
+ * It is the command when the user inputs a bye
+ */
+public class CommandBye extends Command {
+
+    public CommandBye(TaskList taskList, Ui ui) {
+        super(taskList, ui);
+    }
+
+    /**
+     * {@inheritDoc}
```
Kappaccinoh: cool - I didnt know you could do this

Kappaccinoh (commented):
Great job on the iP thus far, it seems like you have covered the tasks needed. Overall no complaints about coding standards. You have added javadocs, named your variables correctly, organised your files in a manner that is noticeably consistent (and is hence readable in quite a quick fashion because I know where to look), and the names of the classes are also well named and as a result self explanatory!

PR #215, opened 2024-01-30
[qinboan] iP

qinboan:
DiveBot frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- FAST SUPER FAST to use

All you need to do is,

2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

```ruby
public class Main {
    }
}
```


src/main/java/Deadline.java
```diff
@@ -0,0 +1,31 @@
+import java.time.LocalDate;
```
JerryO3: You may want to package your classes so that contributors are better able to understand the code base!

src/main/java/Deadline.java
```diff
@@ -0,0 +1,31 @@
...
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
```
JerryO3: Java Documentation here will be good to explain the functionality of the class.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,31 @@
...
+public class Deadline extends Task {
+    private LocalDate byDate;
+
```
JerryO3: Also a good idea to provide Java Documentation for constructors, especially since you have exception handling logic that makes this method non-trivial 😄

src/main/java/Duke.java
```diff
@@ -1,10 +1,241 @@
...
+    private static void markTask(String input) throws DukeException {
+        int taskIndex = Integer.parseInt(input.substring(5)) - 1;
+        if (taskIndex >= 0 && taskIndex < tasks.size()) {
+            tasks.get(taskIndex).markAsDone();
+            System.out.println("Nice! I've marked this task as done:");
+            System.out.println(tasks.get(taskIndex).toString() + "\n");
+            saveTasksToFile();
+        } else {
+            throw new DukeException("Task does not exist.");
+            //System.out.println("Task does not exist.");
```
JerryO3: You can consider deleting this commented code segment, and instead use version control functionality to store this code as a previous commit/other branch. Would make your code neater too!

src/main/java/Duke.java
```diff
@@ -1,10 +1,241 @@
+import java.util.Scanner;
```
JerryO3: As mentioned in another comment, would be good if you can provide Java Documentation for all your non-trivial methods and class so that other developers are able to better understand your codebase.

src/main/java/Event.java
```diff
@@ -0,0 +1,42 @@
+import java.time.LocalDate;
```
JerryO3: Same comments about Javadocs headers

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
+public class Task {
```
JerryO3: Missing Javadocs headers, mentioned in another comment.

JerryO3 (commented):
Would be good if you could provide Java Documentation for your classes and methods for others' reference! Also, each class should ideally be packaged for easier handling.

src/main/java/Task.java
kervyntan: Missing Javadocs headers

src/main/java/Task.java
```diff
@@ -0,0 +1,25 @@
...
+    public void markAsDone() {
+        this.isDone = true;
+    }
+
+    public void markAsNotDone() {
+        this.isDone = false;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "[X]" : "[ ]");
```
kervyntan: Parantheses not needed around this expression!

kervyntan (commented):
Overall good job! Just generally small things to take note of

PR #220, opened 2024-01-30
[yorklim] iP

yorklim:
# NotDuke

>"Your mind is for having ideas, not holding them." – David Aleen ([source](https://dansilvestre.com/productivity-quotes))

NotDuke frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- ~~FAST~~ *SUPER FAST* to use

All you need to do is,

1. download it from [here](https://github.com/yorklim/ip/releases).
1. double-click it.
1. add your tasks.
1. let it manage your tasks for you :wink:

And its **FREE**!

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)

---
```
public class Main {
    }
}
```


src/main/java/duke/command/ByeCommand.java
```diff
@@ -0,0 +1,17 @@
+package duke.command;
+
+import duke.dukeexception.DukeException;
+import duke.storage.Storage;
+import duke.tasklist.TaskList;
+import duke.ui.Ui;
+public class ByeCommand extends Command {
```
ongzhili: I like that the imports are sorted alphabetically. However, would it be better to leave a line after line 6 to improve readability?

src/main/java/duke/parser/Parser.java
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Returns the corresponding command given the input string.
+     * @param fullCommand the String command to be parsed
+     * @return the command corresponding to the string
+     * @throws DukeInvalidCommand if string is not a valid command
+     */
+    public static Command parse(String fullCommand) throws DukeInvalidCommand {
+        String[] input = fullCommand.split(" ", 2);
+        String command = input[0];
+        if (command.equals("bye")) {
```
ongzhili: I like that the if else statements are formatted nicely, but I would also like to suggest switch case as a slightly less verbose alternative.

src/main/java/duke/tasklist/TaskList.java
```diff
@@ -0,0 +1,155 @@
...
+    public String add(Task newTask) {
+        list.add(newTask);
+        return "Got it. I've added this task:" + System.lineSeparator() + newTask.taskInfo() + listStatus();
+    }
+
+    /**
+     * Returns a string to be printed which contains all the String representation of the tasks in the TaskList.
+     * @return a string representation of all the task in the TaskList.
+     */
+    public String list() {
```
ongzhili: Since there is a list interface already in Java, should this method name be changed to prevent confusion?

src/main/java/duke/tasklist/TaskList.java
```diff
@@ -0,0 +1,155 @@
...
+
+import duke.dukeexception.DukeException;
+import duke.dukeexception.DukeTaskNotFound;
+import duke.task.Task;
+
+/**
+ * A List that is used to contain the Tasks inputted by the user.
+ */
+public class TaskList {
+    private ArrayList<Task> list;
```
ongzhili: Since there is a list interface already in Java, should this variable name be changed to prevent confusion?

src/main/java/duke/command/CheckCommand.java
```diff
@@ -0,0 +1,31 @@
+package duke.command;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
```
ongzhili: I like the separation of import statements. Common mistake people leave out.

src/main/java/duke/task/Events.java
```diff
@@ -0,0 +1,68 @@
...
+        super(name, status);
+        this.from = from;
+        this.to = to;
+
+    }
+
+    @Override
+    public String happenOn(LocalDate date) {
+        if ((date.isAfter(from.toLocalDate()) && date.isBefore(to.toLocalDate()))
+                || date.isEqual(from.toLocalDate()) || date.isEqual(to.toLocalDate())) {
```
ongzhili: I like the clean like break.

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,108 @@
...
+     */
+
+    public String taskInfo() {
+        String output = "";
+        if (isDone) {
+            output += "[X]";
+        } else {
+            output += "[ ]";
+        }
+        return output + " " + name;
```
ongzhili: Could this be shortened to the following?
```suggestion
        if (isDone) {
            return "[X] " + name;
        } else {
            return "[ ] " + name;
        }
```

ongzhili (commented):
Looks good 👍 .
Some small changes suggested.

PR #223, opened 2024-01-30
[Tang Yetong] iP

darkv32:
# Kewgy

### **Simple** chatbot made by me! Advantages:
- Simple
- Chatbot
- Easy to use (simple)

### Uses _file IO_ to make it better:

1. Store
2. Load
3. Save

### Uses OOP concepts, here is a snippet from the `main` method:

```java
public class Duke {

        Storage storage = new Storage();
        Ui ui = new Ui();
        Parser parser = new Parser();
        List<Task> userTaskList = storage.loadTasks();
		...
	}
}
```

### TODO

- [X] CLI
- [ ] JavaFX

### An emoji

👻

### Block quote

> Here is a block quote

### Download

Download it from [here!](https://github.com/darkvoid32/ip/releases/tag/A-Jar)

###


src/main/java/Tasks/Task.java
```diff
@@ -0,0 +1,53 @@
...
+                    return T;
+                case "D":
+                    return D;
+                case "E":
+                    return E;
+                default:
+                    return null;
+            }
+        }
+    }
```
LimZiJia: Is this really necessary? Right now, all you use it for is to print the letters. I think it adds very little because the enum is not descriptive enough. For example instead of ```TaskType.T```, I think ```TaskType.ToDo``` would be more appropriate.

In the current iteration, I think even just replacing ```TaskType.X``` in the subclasses with the corresponding ```"X"``` might be better.

src/main/java/Tasks/Deadline.java
```diff
@@ -0,0 +1,50 @@
...
+            super.description = descParsed[0].trim() + " (by: " + formattedDate + ")";
+        } catch (Exception e) {
+            // System.out.println("Unknown Date Format!");
+            super.description = descParsed[0].trim() + " (by: " + descParsed[1].trim() + ")";
+        }
+    }
+
+    @Override
+    public String toString() {
+        return "[" + TaskType.D + "][" + super.getStatusIcon() + "] " + super.description;
```
LimZiJia: Just a suggestion but you could explore using ```String.format()``` to make your code more readable. It is not that bad now because it is still short.
```suggestion
        return String.format("[%s][%s] %s\n", TaskType.D, super.getStatusIcon(), super.description);
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,138 @@
...
+        ui.addConversation(userText, dukeText);
+    }
+
+    public String getResponse(String userMsg) {
+        String[] userMsgParsed = userMsg.split(" ", 2);
+        Command nextCommand = parser.parseUserMsg(userMsgParsed);
+
+        String dukeText = "";
+
+        switch (nextCommand) {
```
LimZiJia: Very nice use of enums and switch statements. This will be my inspiration to implement enums. Do note that the ```case``` and ```default``` statements should have the same indentation as the switch as per coding standard.

src/main/java/Duke.java
```diff
@@ -1,10 +1,138 @@
...
+                    dukeText = ui.printAddTask(task, userTaskList.size());
+                } catch (DukeException e) {
+                    dukeText = ui.printError(e);
+                }
+                break;
+            case DEADLINE:
+                try {
+                    Deadline task = new Deadline(userMsgParsed[1]);
+                    userTaskList.add(task);
+                    storage.saveTask(userMsgParsed[1], TaskType.T);
```
LimZiJia: There is a bug here!
```suggestion
                    storage.saveTask(userMsgParsed[1], TaskType.D);
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,138 @@
...
+                    dukeText = ui.printAddTask(task, userTaskList.size());
+                } catch (DukeException e) {
+                    dukeText = ui.printError(e);
+                }
+                break;
+            case EVENT:
+                try {
+                    Event task = new Event(userMsgParsed[1]);
+                    userTaskList.add(task);
+                    storage.saveTask(userMsgParsed[1], TaskType.T);
```
LimZiJia: Another bug here from ctrl + c 😆 
```suggestion
                    storage.saveTask(userMsgParsed[1], TaskType.E);
```

src/main/java/Ui/Ui.java
```diff
@@ -0,0 +1,202 @@
...
+            if (userTaskList.get(i).getDescription().contains(keyword)) {
+                tasks += i + ": " + userTaskList.get(i) + "\n";
+            }
+        }
+        
+        return this.formatString(
+            "Here are the matching tasks in your list:\n" + 
+            tasks + "\n"
+        );
+    }
```
LimZiJia: I think with a more complicated function for ```userTaskList``` like this, it would be clearer to create a class ```TaskList```, so that in the future, we can easily add functions to the ```userTaskList``` instead of letting the UI complete this function.

LimZiJia (commented):
Overall great readability. There is one use of enums that I will take inspiration from and another that is not that useful (```TaskType```). Great work! 👍

src/main/java/Duke.java
darkv32: Good eye, thanks for finding it 😄

PR #230, opened 2024-01-30
[ReganChoy] iP

ReganChoy:
# DukeXAerial

>Time is of the Essence

- Minimal Crashes
* FAST

1. Early Access 🧣 
2. ~~Lifetime Support~~

Some application commands include:
```
todo <Task>
deadline <Task> dd/mm/yyyy hh:mm
bye
list
mark <Task Number>
```

- [x] Early build
- [ ] Full Test

>[!CAUTION]
>DO NOT AMEND THE TXT FILE TO PREVENT DATA CORRUPTION

[Main Project Source](https://nus-cs2103-ay2324s2.github.io/website/admin/index.html)


docs/README.md
```diff
@@ -1,29 +1,30 @@
-# User Guide
+# Duke User Guide
```
solomonng2001: why not call it Gundam? 😜

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,63 @@
...
+     */
+    public String getPeriod() {
+        return " (By: " + dateToString() + ")";
+    }
+
+    /**
+     * Informs the user a new Deadline has been created.
+     * @return String informing a new Deadline has been created.
+     */
+    public String announcement() {
```
solomonng2001: Can this be renamed to an action? getAnnouncment etc? I noticed the announcement() method in other files as well

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,63 @@
...
+        return "New Deadline created!";
+    }
+
+    /**
+     * Returns a string representation of the Deadline for txt file saving purposes.
+     * @return String representation of the Deadline for txt file saving purposes.
+     */
+    public String saveString() {
```
solomonng2001, line 51: Is there an easier way other than using multiple concatenation operators? Would the following work for you? Actually, I think this is personally preference, but would like to let you know of alternatives. Concatenation could be more readable too, since variables and string can be read sequentially.

Screenshot 2024-02-07 at 7 07 14 PM

I noticed similar instances elsewhere.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,102 @@
...
+     * @param commandInput
+     * @param duke
+     * @param taskList
+     */
+    public String input(String commandInput, Duke duke, TaskList taskList) {
+        boolean isCommandValid = false;
+        this.command = "";
+        this.secondaryInput = "";
+
+        if (commandInput.equals("save")) {
```
solomonng2001, line 30: Is there a way to keep track of loose strings like "save" in a constant or enum? A central area to store such strings as constants can make a large codebase more readable. The benefits are as follows
1. can go to a command enum file or command constants to see all types of commands
2. can change all instances of the same "save" command, by changing the string saved in a constant or enum to "update"

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,110 @@
...
+     */
+    public String announcement() {
+        return "New task created!";
+    }
+
+    /**
+     * Method generates the string of the task in format for file txt saving.
+     * @return String representation of the task in the format for file txt saving.
+     */
+    public String saveString() {
```
solomonng2001, line 99: Should this be saved as a constant (static final) instead, since it does not operate on the string but only returns it as it is?

Noticed similar instances elsewhere, eg with announcment()

solomonng2001 (commented):
Good job Regan, just some nits.

Strong OOP and good handling of exceptions. I focused more on syntax and standards, rather than functional. Hope they are useful. Additionally, some of the methods could be made private without changing functionality of the code, but up to you if you would like to implement it that way.

Overall great job! 😄

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,102 @@
...
+
+    }
+
+    /**
+     * Method is responsible for interpreting commands send by the user and handling invalid commands.
+     * @param commandInput
+     * @param duke
+     * @param taskList
+     */
+    public String input(String commandInput, Duke duke, TaskList taskList) {
```
jimseah-0116: Since this is a method, following the coding convention, maybe you could give the method a verb name instead?
For example, public String parseInput(...) could work.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,83 @@
...
+        file.getParentFile().mkdirs();
+
+        if (!file.exists()) {
+            file.createNewFile();
+        }
+
+        Scanner scanner = new Scanner(file);
+        while (scanner.hasNext()) {
+            String inputLine = scanner.nextLine();
+            String[] inputComponents = inputLine.split("\\|", 5);
```
jimseah-0116, line 41: Variable names are in camelCase. Great!

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,76 @@
...
+public class Event extends Task {
+    private LocalDateTime start;
+    private LocalDateTime end;
+
+    /**
+     * Constructor for Event Class.
+     * @param task
+     * @param taskType
+     * @param start
+     * @param end
```
jimseah-0116, line 19: Perhaps these @param tags may be unnecessary?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,166 @@
...
+
+        reply += "We have " + (taskList.size()) + " tasks.\n";
+
+        return reply;
+    }
+
+    /**
+     * Adds new task to the TaskListt.
+     * @param first
+     * @param second
```
jimseah-0116: The parameter names are a bit confusing - perhaps you may want to specify that 'first' is the Task type and second is the rest of the String?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,58 @@
...
+ * Duke Class is the main class that will run based on the different commands given
+ * by the user. Available commands include todo, event, deadline, mark, unmark, delete.
+ */
+public class Duke {
+    private Storage storage;
+    private TaskList taskList;
+    private Parser parser = new Parser();
+    private Scanner scanner;
+    private boolean isEnded = false;
+    private final String PATH = "data/tasks.txt";
```
jimseah-0116, line 18: Constant PATH is in all upper case letters. Great!

src/main/java/duke/DialogBox.java
```diff
@@ -0,0 +1,61 @@
...
+import javafx.scene.image.Image;
+import javafx.scene.image.ImageView;
+import javafx.scene.layout.HBox;
+
+/**
+ * A custom control using FXML.
+ * This control represents a dialog box consisting of an ImageView to represent the speaker's face and a label
+ * containing text from the speaker.
+ */
+public class DialogBox extends HBox {
```
jimseah-0116, line 22: Class name is PascalCase. Great!

src/main/java/duke/Deadline.java
ReganChoy: Good idea!

src/main/java/duke/TaskList.java
ReganChoy: Good suggestion!

PR #234, opened 2024-01-30
[Tan Zi Yang] iP

ziyang27:
# Dibo the Task Chatbot 

Dibo frees your mind from having to remember things you need to do. It's
- test-based
- easy to learn
- ~~FAST~~ _SUPER FAST_ to use

> [!IMPORTANT]
> Take a look at the list of commands and their behaviors to familiarise yourself for **MAXIMUM SPEED!**

All you need to do is,
1. download it from [here](https://github.com/ziyang27/ip).
2. double-click it.
3. add your tasks.
4. sit back and let it manage your task for you :100:

Most importantly, it is totally **FREE!**

Key features:
- [x] Managing tasks
- [x] Saving tasks to harddrive
- [ ] Managing deadlines
- [ ] Reminders (coming soon)

***
Before we begin, let's look at some of the ***COOL*** 😎 commands!
- `list` lists all the tasks in the task list, their completion status, type, and other relevant information.
- `mark` and `unmark` to mark the respective task as done or not.
- `delete` deletes the task from the task list.

***
In case you are wondering, here's the main method:
```java
        Dibo dibo = new Dibo("./data/dibo.txt");
        dibo.run();
}
```
**What? Just like that?** 
Yes, click in to find out more! :monocle_face:


src/main/java/dibo/Dibo.java
```diff
@@ -0,0 +1,47 @@
...
+        try {
+            this.tasks = new TaskList(this.storage.loadData());
+        } catch (DiboException e) {
+            this.ui.showError(e.getMessage());
+            this.tasks = new TaskList(new ArrayList<>());
+        }
+    }
+
+    public static void main(String[] args) {
+        Dibo dibo = new Dibo("./data/dibo.txt");
```
SherisseTJW: It might be better to store the filepath as a private final field

src/main/java/dibo/Parser.java
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * Takes in the command and tries to make sense of it.
+     * Returns a Command object which can be run at the end.
+     *
+     * @return A Command object to be run.
+     * @throws DiboException If an error occurs when parsing.
+     */
+    public static Command parse(String fullCommand) throws DiboException {
+        String[] commandDetails = fullCommand.split(" ");
+        switch (commandDetails[0]) {
```
SherisseTJW, line 35: Maybe you can add the `break;` statement for all the cases or have a `// Fallthrough` comment?

src/main/java/dibo/Parser.java
```diff
@@ -0,0 +1,146 @@
...
+        String startString = details[1].substring(4).trim();
+        return LocalDate.parse(startString, Parser.INPUT_FORMAT);
+    }
+
+    private static LocalDate getEndDate(String fullCommand)
+            throws DateTimeParseException, IndexOutOfBoundsException {
+        String[] details = fullCommand.split("/");
+        String endString = details[2].substring(2).trim();
+        return LocalDate.parse(endString, Parser.INPUT_FORMAT);
+    }
```
SherisseTJW: Not sure I like how these helper methods are `private static` rather than just `private`. Is there a rationale behind this?

src/main/java/dibo/Storage.java
```diff
@@ -0,0 +1,123 @@
...
+                    String description1 = details[2];
+                    task = new ToDo(description1);
+                    break;
+                case "deadline":
+                    String description2 = details[2];
+                    String by = details[3].trim();
+                    task = new Deadline(description2, convertToLocalDate(by));
+                    break;
+                case "event":
+                    String description3 = details[2];
```
SherisseTJW: Not sure I like the name `description1`, perhaps you can rename it to `todoDescription` instead? Or have a singular `description` declaration at line 50 above which can be used for each of the cases since you have `description2` and `description3` as well?

src/main/java/dibo/TaskList.java
```diff
@@ -0,0 +1,144 @@
...
+    /**
+     * Returns the string representation of the list to be displayed.
+     */
+    public String getDisplayFormat() {
+        StringBuilder list = new StringBuilder("Here are the tasks in your list:\n");
+        for (int i = 0; i < count; ++i) {
+            list.append((i + 1));
+            list.append(".");
+            list.append(tasks.get(i).toString());
+            list.append("\n");
```
SherisseTJW: Perhaps you can consider using `String.format()` to have it in 1 line?

SherisseTJW (commented):
Overall, LGTM!

Just some minor suggestions to clean up the code ^ ^

src/main/java/dibo/Parser.java
ziyang27: Actually, I didn't take notice of this before. I will take note in the future! Thank you so much 😄

src/main/java/dibo/TaskList.java
ziyang27: Thank you for the suggestion! I will look into it!

JonChong98 (commented):
Can't find any glaring issues with code quality aside from the minor ones already highlighted by fellow reviewer. Good job!

PR #248, opened 2024-01-30
[owenong1] iP

owenong1:
# BotBot
> “Your mind is for having ideas, not holding them.” – David Allen ([source](https://dansilvestre.com/productivity-quotes))

BotBot frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- FAST SUPER FAST to use
All you need to do is,

2. double-click it.
3. add your tasks.
And it is **FREE!**

Features:

- [x] Managing tasks
- [ ] Managing deadlines 
 - [ ] Reminders (coming soon)


src/test/java/botbot/task/TaskTest.java
```diff
@@ -0,0 +1,20 @@
+package botbot.task;
+
+import botbot.exception.InvalidDateException;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class TaskTest {
+    @Test
+    public void getRepToDoTest() {
```
ibnu2651, line 10: I think for tests the naming convention is usually testToDoGetRep or something like that?

src/main/java/botbot/task/Task.java
```diff
@@ -0,0 +1,75 @@
+package botbot.task;
+
+import botbot.exception.InvalidDateException;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+public class Task {
+    String taskName;
+    Boolean done = false;
```
ibnu2651, line 9: I'm not sure if this is within the scope of naming but I think for Java, Boolean is actually the wrapper class for the primitive value boolean which might be more suitable here? But if you meant to use the wrapper class then by all means.

src/main/java/botbot/exception/DescriptionException.java
```diff
@@ -0,0 +1,7 @@
+package botbot.exception;
+
+public class DescriptionException extends BotBotException {
```
ibnu2651, line 3: I think having a more descriptive name would help since DescriptionException does not really say what is wrong, maybe something like MissingDescriptionException would more clearly capture the nature of the error.

src/main/java/botbot/task/Task.java
```diff
@@ -0,0 +1,75 @@
...
+        return this.taskName;
+    }
+    public void mark() {
+        this.done = true;
+    }
+    public void unmark() { this.done = false; }
+
+    /**
+     * Returns string format of a task
+     * @return String rep
```
ibnu2651, line 31: While I think it's ok to use "rep" instead of "representation" for the method names, maybe you can use the full word for the Javadoc since it's supposed to describe the method in full.

src/main/java/botbot/BotBot.java
```diff
@@ -0,0 +1,40 @@
...
+
+public class BotBot {
+    private Storage storage;
+    private TaskList taskList;
+    private Ui ui;
+    public static String LIST_PATH = "data/list.txt";
+
+    /**
+     * Returns a new instance of chatbot
+     * @throws BotBotException
```
ibnu2651, line 18: I think there should be an empty line between the method description and the params/throws/returns.

ibnu2651 (commented):
I think this is a pretty good effort that covers all the requirements. I was more concerned with the naming aspect of code quality and I could not find much fault so I raised some other concerns. Keep up the good work!

src/main/java/botbot/Ui.java
vannykin, line 1: Good naming of all methods in all classes, very clean if, else-if, else segments, and clear logic.
Great readability! I don't see any coding standard violations :) yay

vannykin (commented):
Good job :) I should probably go look through mine now and improve my own readability

PR #250, opened 2024-01-30
[Timothy Soh] iP

LargeCrowd:
# CowboyBebot (aka Bebot)
> "Its just a chatbot"

Bebot frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn

All you need to do is,

1. download it from [NOT here](https://nus-cs2103-ay2324s2.github.io/website/schedule/week4/project.html).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [X]  Managing tasks
- [X]  Managing deadlines


src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,40 @@
...
+        this.byString = by;
+        try {
+            this.byDate = parseDate(by, f1, f2, f3, f4); // Parse the by string
+        } catch (DateTimeException e) {
+            throw new DateTimeException("Error parsing date in Deadline: " + e.getMessage());
+        }
+    }
+
+    @Override
+    protected String taskType() {
```
dillontkh: We should use verbs for method names in accordance to the [CS2103 Java Coding Standard](https://se-education.org/guides/conventions/java/intermediate.html#naming). I noticed this issue elsewhere too.

```suggestion
    protected String getTaskType() {
```

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * Task class
+ */
+public class Task {
+    protected String task;
+    protected boolean isDone;
+    protected DateTimeFormatter f1 = DateTimeFormatter.ofPattern("d/M/yyyy", Locale.ENGLISH);
+    protected DateTimeFormatter f2 = DateTimeFormatter.ofPattern("yyyy-M-d", Locale.ENGLISH);
+    protected DateTimeFormatter f3 = DateTimeFormatter.ofPattern("d MMM yyyy", Locale.ENGLISH);
+    protected DateTimeFormatter f4 = DateTimeFormatter.ofPattern("d-M-yyyy", Locale.ENGLISH);
```
dillontkh: Perhaps use clearer variable names? `f1` to `f4` doesn't convey much meaning 😅

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,41 @@
...
+public class Parser {
+    private Ui ui;
+    private TaskList taskList;
+
+    public Parser(TaskList taskList) {
+        this.ui = new Ui();
+        this.taskList = taskList;
+    }
+
+    public boolean parseInput(String input) {
```
dillontkh: Maybe the method name could be more descriptive? 

It appears that you are handling the detection and execution of the command here. Maybe something like `executeUserInput()`?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,111 @@
...
+
+                        Events newEvent = new Events(eventDescription, eventFrom, eventTo);
+                        if (taskStatus.equals("done")) {
+                            newEvent.markDone();
+                        }
+                        this.taskList.add(newEvent);
+                        break;
+                    default:
+                        System.out.println("Unrecognized task type: " + task);
+                        throw new IOException("Error loading data from file: unrecognized task type");
```
dillontkh: Should `case` statements be indented according to the [CS2103 Java Coding Standards](https://se-education.org/guides/conventions/java/intermediate.html#layout)?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,238 @@
...
+        } else if (task.startsWith("deadline")) {
+            return TaskType.DEADLINE;
+        } else if (task.startsWith("event")) {
+            return TaskType.EVENT;
+        } else if (task.equals("help")) {
+            return TaskType.HELP;
+        } else if (task.startsWith("delete")) {
+            return TaskType.DELETE;
+        } else {
+            return TaskType.UNKNOWN;
```
dillontkh: Was there a reason you chose to explicitly include an `UNKNOWN` task type? I thought the `default` part of your `switch` statement would have already caught any invalid commands.

dillontkh (commented):
Overall, code was easy to understand. Variables and methods were given descriptive names apart from a few minor exceptions. You might want to consider using checkstyle to catch any CS2103 Java Coding Standard violations. Keep up the good work!

src/main/java/duke/Task.java
LargeCrowd: it felt neater to give ` f1` rather than `format1`, which is why i made this choice, do lmk if you still think its more appropriate to do `format1` though

src/main/java/duke/TaskList.java
LargeCrowd: youre right thats an oversight on my part

src/main/java/duke/Task.java
dillontkh: As someone who has recently done this increment I know it's because they asked us to accept and display the dates in different formats, but to someone who isn't familiar with the application requirements this might be a little confusing.

What I did was name them something like `UserInputFormat`, `DisplayFormat`, `SaveFormat` etc. This way, it provides a little bit more information on the intended usage of these formats rather than just `format1`, `format2`, etc.

LargeCrowd: okay gets. actually i just put them all in as like kinda edge cases of other date formats being input. like giving it more options to not throw an exception but okay ill see how i can rename them a bit better

PR #257, opened 2024-01-30
[myang2020] iP

myang2020:
### Duke - Dwight Schrute: A tasks manager for fellow "The Office" Fans

> “Whenever I’m about to do something, I think, ‘Would an idiot do that?’ and if they would, I do not do that thing.” - Dwight Schrute

Duke - Dwight Schrute gives you a peace of mind by helping you keep track of tasks. It is -

- text-based
- complete with easy commands

To download:

download the zip folder from https://github.com/nus-cs2103-AY2324S2/ip.

1. double-click it
2. start using
3. And it is ~~ expensive~~FREE!

Features:

- [X]  Managing tasks 😆

If you are a Java programmer, here is the main method
```
    public static void main(String[] args) throws IOException, DukeException {
        new Duke("./data/duke.txt").run();
    }
```


src/main/java/duke/util/Ui.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public final String divider = "---------------------------------------------------------------";
+    public Ui() {
+    }
+
+    /**
+     * Prints welcome message
+     */
+    public void welcome() {
+        String chat_name = "Dwight Schrute";
```
Apzure: Intimidating name for chatbot

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,168 @@
...
+                storage.writeFile(todo.getList());
+            } catch (DukeException err) {
+                System.out.println(err.getMessage());
+                System.out.printf("\t%s\n", divider);
+            } catch (IOException err) {
+                System.out.println(err.getMessage());
+                System.out.printf("\t%s\n", divider);
+            }
+        }
+    }
```
Apzure: 1. It might be better to have a separate class to handle commands.
2. Abstract out the printing of tasks-related messages to the methods inside Task.java themselves.

src/main/java/duke/task/EventTask.java
```diff
@@ -0,0 +1,43 @@
...
+
+    }
+
+    /**
+     * String representation of EventTask
+     *
+     * @return String representation of EventTask
+     */
+    @Override
+    public String toString() {
```
Apzure: Implement the toString() method in your Task.java.
This way you can call it by super.toString() in each of the subclasses of Task.java.
This is helpful since each of the subclasses toString() methods have commonalities in the string it prints.
For reference, this is mine:
 String check = isDone ? "[X] " : "[ ] ";
 return check + this.name;

src/main/java/duke/util/Storage.java
```diff
@@ -0,0 +1,92 @@
...
+                            break;
+                        }
+
+                    case ("deadline"):
+                        first_string = input.split(" /")[0];
+                        second_string = input.split(" /")[1];
+                        task_name = String.join(" ", Arrays.copyOfRange(first_string.split(" "), 1, first_string.split(" ").length));
+                        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("d/M/yyyy HHmm");
+                        end = String.join(" ", Arrays.copyOfRange(second_string.split("by "), 1, second_string.split("by ").length));
+                        LocalDateTime end_time = LocalDateTime.parse(end, formatter);
```
Apzure: More readable if you abstract this into a method

Apzure (commented):
LGTM! Keep up the good work!
You can consider some of the changes I recommended!

PR #258, opened 2024-01-30
[1rbg] iP

1rbg:
CoDriver is a new chatbot, running GPT(-1). It's 
- text-based
- easy to learn
- ~~FAST~~ _Super_ FAST to use

All you need to do is,
1. download it from [here](https://github.com/1rbg/ip/releases/tag/A-Jar).
2. launch it in terminal using `java -jar codriver.jar`.
3. add your tasks.

And it is **FREE**!

Features:
- [x] Managing tasks
- [ ] Reminders (coming soon)

------------------------------------------------------------
``` Java
public class Main {
	    new CoDriver("./data/codriver.txt").run();  
	}
}
```


src/main/java/lrbg/codriver/data/Task.java
```diff
@@ -0,0 +1,64 @@
...
+    @Override
+    public String toString() {
+        return "[" + getStatusIcon() + "] " + this.description;
+    }
+
+    /**
+     * Returns a string representation of the task to be saved in a file.
+     * @return The string representation of the task to be saved.
+     */
+    public String toSaveString() {
```
PallonCX: The lecture topic for Week 4 content suggests that:

> A name is not just for differentiation; it should explain the named entity to the reader accurately and at a sufficient level of detail.

Considering the method in line 52 is called `toString`, which generates a `String` as its return value, a reader may naturally interpret it as **a function that describes the object as text**.

However, the method name for line 60 is called `toSaveString`. A reader may not have a clear idea about "what a SaveString is", or may interpret the function as **a function that saves a string somewhere**. These two interpretations are clearly incorrect if the reader reviews the method implementation.

Therefore as a suggestion, perhaps it will be good to give the function a better name? One of the example that comes to my mind is `toStringForSaveInFile`.

src/main/java/lrbg/codriver/data/TaskList.java
```diff
@@ -0,0 +1,170 @@
...
+                LocalDate eventFrom = Parser.parseDate(dates[0]);
+                LocalDate eventTo = Parser.parseDate(dates[1]);
+                Event e = new Event(eventDescription, eventFrom, eventTo);
+                if (isDone) {
+                    e.markDone();
+                }
+                this.tasks.add(e);
+                break;
+            }
+        }
```
PallonCX, line 63: The lecture topic for Week 4 content suggests that:

> A name is not just for differentiation; it should explain the named entity to the reader accurately and at a sufficient level of detail.

Noticed that within these part of codes, you have used `s`, `t`, `d`, and `e` as your variable name. Perhaps name formed by single character is more suitable for temporary usage such as loop variable, but for here I think giving a more complete variable name can effectively improve the readability of code.

src/main/java/lrbg/codriver/command/ByeCommand.java
```diff
@@ -0,0 +1,40 @@
...
+import lrbg.codriver.ui.Ui;
+
+/**
+ * Represents a command to exit the program.
+ */
+public class ByeCommand extends Command {
+    /**
+     * Creates a new ByeCommand.
+     */
+    public ByeCommand() {
```
meenulekha, line 15: I think the default constructor will be used even if you don't explicitly include one!

src/main/java/lrbg/codriver/parser/Parser.java
```diff
@@ -0,0 +1,195 @@
...
+            }
+            return new TodoCommand(commandLine.substring(5));
+        case "deadline": {
+            if (arguments.length < 2) {
+                throw new CoDriverException("Error! You cannot provide a deadline with no parameters!");
+            }
+
+            StringBuilder descriptionBuilder = new StringBuilder();
+            int i;
+            for (i = 1; i < arguments.length; i++) {
```
PallonCX: The Java Coding Standard that we are referring suggests that:

> Variables should be initialized where they are declared and they should be declared in the smallest scope possible.

Hence, perhaps you may want to modify the code as such:
```Suggestion
            int i = 1;
            for (; i < arguments.length; i++) {
```

src/main/java/lrbg/codriver/parser/Parser.java
```diff
@@ -0,0 +1,195 @@
...
+            if (descriptionBuilder.length() == 0) {
+                throw new CoDriverException("Error! You cannot provide an event with no description!");
+            }
+            descriptionBuilder.deleteCharAt(descriptionBuilder.length() - 1); // remove the last space
+
+            if (i >= arguments.length - 1) { // if the last word is /from or there is no /from
+                throw new CoDriverException("Error! You must provide a /from date for an event!");
+            }
+
+//        StringBuilder fromBuilder = new StringBuilder();
```
meenulekha: make sure your comments are aligned with the rest of the code:)

src/main/java/lrbg/codriver/parser/Parser.java
```diff
@@ -0,0 +1,195 @@
...
+            for (; i < arguments.length; i++) {
+                if (arguments[i].equals("/to")) {
+                    break;
+                }
+                // check if in yyyy-mm-dd format
+                fromDate = Parser.parseDate(arguments[i]);
+                if (fromDate == null) {
+                    throw new CoDriverException("Error! The date provided must be in yyyy-mm-dd format!");
+                }
+//            fromBuilder.append(arguments[i]).append(" ");
```
meenulekha: you can remove any comments of code that is no longer being used :)

src/main/java/lrbg/codriver/parser/Parser.java
```diff
@@ -0,0 +1,195 @@
...
+                if (arguments[i].equals("/from")) {
+                    break;
+                }
+                descriptionBuilder.append(arguments[i]).append(" ");
+            }
+            if (descriptionBuilder.length() == 0) {
+                throw new CoDriverException("Error! You cannot provide an event with no description!");
+            }
+            descriptionBuilder.deleteCharAt(descriptionBuilder.length() - 1); // remove the last space
+
```
PallonCX, line 95: The Java Coding Standard that we are referring suggests that:

> Logical units within a block should be separated by one blank line.

Hence, perhaps you may want to insert some blank lines to improve the code readability?

PallonCX (commented):
Well done! LGTM. Just some nits to fix.

src/main/java/lrbg/codriver/command/Command.java
```diff
@@ -0,0 +1,30 @@
...
+     * @throws CoDriverException If an error occurs during execution.
+     */
+    public abstract void execute(TaskList tasks, Ui ui, Storage storage) throws CoDriverException;
+
+    /**
+     * Returns true if the command is an exit command.
+     *
+     * @return True if the command is an exit command.
+     */
+    public boolean isExit() {
```
tanguoyang, line 28: As the list of commands grow, there will only ever be 1 exit command. I wonder if theres a better way to structure the code instead of having every command have isExit() = false?

src/main/java/lrbg/codriver/command/DeadlineCommand.java
```diff
@@ -0,0 +1,55 @@
...
+     */
+    public DeadlineCommand(String description, LocalDate date) {
+        this.description = description;
+        this.date = date;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws CoDriverException {
```
tanguoyang: I wonder if theres a better name for CoDriverException? CoDriverException doesnt tell me much about what kind of error is happening

src/main/java/lrbg/codriver/command/DeleteCommand.java
```diff
@@ -0,0 +1,48 @@
...
+        tasks.deleteTask(index);
+        ui.showDeleteTask(task, tasks.size());
+    }
+
+    /**
+     * Returns true if the given object is equal to this command, only for testing purposes.
+     * @param obj The object to compare to.
+     * @return True if the given object contains the same index, false otherwise.
+     */
+    public boolean testEquals(Object obj) {
```
tanguoyang, line 38: consider overriding equals() and hashcode() function instead!

src/main/java/lrbg/codriver/command/EventCommand.java
```diff
@@ -0,0 +1,40 @@
...
+        this.toDate = toDate;
+    }
+
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws CoDriverException {
+        Task newTask = new Event(this.description, this.fromDate, this.toDate);
+        tasks.addTask(newTask);
+        ui.showAddTask(newTask, tasks.size());
+    }
+
+    public boolean testEquals(Object obj) {
```
tanguoyang, line 32: i wonder if testEquals is better placed somewhere else, like a CommandTest class built for testing instead?

src/test/java/lrbg/codriver/parser/ParserTest.java
```diff
@@ -0,0 +1,158 @@
...
+            assertTrue((new MarkCommand(1)).testEquals(Parser.parse("mark 1")));
+            assertTrue((new UnmarkCommand(1)).testEquals(Parser.parse("unmark 1")));
+            assertTrue((new FindCommand("1234")).testEquals(Parser.parse("find 1234")));
+        } catch (Exception e) {
+            fail();
+        }
+    }
+
+    @Test
+    public void parseTest_deadlineIncorrectArgument_exceptionThrown() {
```
tanguoyang, line 46: Might be better if our deadlineIncorrectArgument has a more specific case. There are many ways in which the argument is incorrect. when i run the test in my terminal, i should be able to know what kind of incorrect argument throws an error!

src/test/java/lrbg/codriver/parser/ParserTest.java
```diff
@@ -0,0 +1,158 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
+
+/**
+ * Test Class for Parser.
+ */
+public class ParserTest {
+    @Test
+    public void parseTest_correctArgument_success() {
```
tanguoyang, line 24: Might be better if our correctArgument has a more specific case. There are many ways in which the argument is correct. when i run the test in my terminal, if it fails i should be able to look at the test name and quickly come up with a diagnosis!

tanguoyang (commented):
minor comments, keep up the good work!

src/main/java/lrbg/codriver/data/TaskList.java
```diff
@@ -0,0 +1,170 @@
...
+                String[] dates = arguments[3].split("~");
+                LocalDate eventFrom = Parser.parseDate(dates[0]);
+                LocalDate eventTo = Parser.parseDate(dates[1]);
+                Event e = new Event(eventDescription, eventFrom, eventTo);
+                if (isDone) {
+                    e.markDone();
+                }
+                this.tasks.add(e);
+                break;
+            }
```
triciiaaa, line 62: Would it be better to include a default branch in `case` statements to ensure that all possible outcomes have been considered at the branching point?

src/main/java/lrbg/codriver/data/TaskList.java
```diff
@@ -0,0 +1,170 @@
...
+        }
+    }
+
+    /**
+     * Gets the size of the list.
+     * @return The size of the list.
+     */
+    public int size() {
+        return this.tasks.size();
+    }
```
triciiaaa: Would it be better to rename the method as `getSize()` instead?

src/main/java/lrbg/codriver/data/TaskList.java
```diff
@@ -0,0 +1,170 @@
...
+
+    /**
+     * Gets the task at the given index.
+     * @param index The index of the task, zero indexed.
+     * @return The task at the given index.
+     */
+    public Task get(int index) {
+        return this.tasks.get(index);
+    }
```
triciiaaa: In addition to the comment on the `size()` method, would it be better to rename this method as `getTask(int index)`?

src/main/java/lrbg/codriver/data/Event.java
```diff
@@ -0,0 +1,39 @@
...
+import java.time.LocalDate;
+
+/**
+ * Represents an event task.
+ */
+public class Event extends Task {
+    /** The start date of the event. */
+    private final LocalDate from;
+    /** The end date of the event. */
+    private final LocalDate to;
```
triciiaaa: Perhaps you may consider renaming the variables `from` and `to` to `startDate` and `endDate` respectively to make it more intuitive for the user?

triciiaaa (commented):
Overall, I found your code easy to read and I like how you separated the different commands into different classes! Keep it up 👍

PR #272, opened 2024-01-30
[yunchao] iP

tanyunchao:
# Martin-the-best-chatbot
> "Coding is hard, but naming variables is harder" [source](chat.openai.com)

## Types of events Martin can handle:
- Todos
- Events
- Deadlines 

Other features:
1. Mark and unmark task 
2. Saves to a **dedicated** file 
3. Able to delete tasks 😄 `delete task 3`

## Heres how to download
- [ ] Download java
- [ ] Fork and download

Run using 
```
java Martin
```


src/main/java/Deadline.java
```diff
@@ -0,0 +1,22 @@
...
+    private final String DATE_TIME_FORMAT = "MMM d yyyy";
+
+    Deadline(String description, LocalDate due) {
+        super(description);
+        this.due = due;
+    }
+
+    @Override
+    public String toString() {
+        return "[D]" + super.toString() + " (by: " + this.due.format(DateTimeFormatter.ofPattern(DATE_TIME_FORMAT)) + ")";
```
teojunda: Perhaps it is more suitable to make the line break before the operator?

src/main/java/Parser.java
```diff
@@ -0,0 +1,49 @@
...
+            case "event":
+                return ChatbotKeyword.EVENT;
+            case "bye":
+                return ChatbotKeyword.BYE;
+            default:
+                throw new IllegalArgumentException("I'm sorry, but I don't know what that means :-(");
+        }
+    }
+
+    public String[] deadlineParse(String input) {
```
teojunda: Perhaps it is better to phrase the method as a verb, like `parseDeadline`?

src/main/java/Parser.java
```diff
@@ -0,0 +1,49 @@
...
+        String[] deadlineArray = input.split(" /by ");
+        String deadlineDescription = deadlineArray[0];
+
+        return new String[]{deadlineDescription, deadlineArray[1]};
+    }
+
+    public String[] eventParse(String input) {
```
teojunda: Perhaps it is better to phrase the method as a verb, like `parseEvent`?

src/main/java/Parser.java
```diff
@@ -0,0 +1,49 @@
...
+            case "todo":
+                return ChatbotKeyword.TODO;
+            case "deadline":
+                return ChatbotKeyword.DEADLINE;
+            case "event":
+                return ChatbotKeyword.EVENT;
+            case "bye":
+                return ChatbotKeyword.BYE;
+            default:
+                throw new IllegalArgumentException("I'm sorry, but I don't know what that means :-(");
```
teojunda: I really like the use of switch case statements and enumerations!

src/main/java/Deadline.java
```diff
@@ -0,0 +1,22 @@
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+public class Deadline extends Task{
```
teojunda: Perhaps you can follow add a space before '{'?

teojunda (commented):
Looks good 👍 
Overall, I like how readable your code was! However, I noticed a few standard coding violations. Also, it adding some JavaDocs to explain the classes and methods would greatly improve the code clarity.

src/main/java/Parser.java
tanyunchao: thanks for raising this up @teojunda , I think its definitely a valid point that i missed

src/main/java/Martin/Parser.java
```diff
@@ -0,0 +1,85 @@
...
+     *
+     * @param input the input string
+     * @return the corresponding ChatbotKeyword
+     * @throws IllegalArgumentException if the command is not recognized
+     */
+    public ChatbotKeyword parse(String input) {
+        String[] inputArray = input.split(" ");
+        String command = inputArray[0].toLowerCase();
+        switch (command) {
+            case "list":
```
chiaryan, line 23: Should the indenting for these `case` blocks be changed to align with the Java [coding standard?](https://se-education.org/guides/conventions/java/intermediate.html#layout)

src/test/java/Martin/ParserTest.java
```diff
@@ -0,0 +1,37 @@
+package Martin;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class ParserTest {
+    @Test
+    public void testGetRemainingWords() {
```
chiaryan: Should this be renamed to follow the naming convention for Junit tests?

src/test/java/Martin/ParserTest.java
```diff
@@ -0,0 +1,37 @@
...
+        Parser parser = new Parser();
+
+        // Test case 1: input with single word
+        String input1 = "Hello";
+        String expectedOutput1 = "";
+        assertEquals(expectedOutput1, parser.getRemainingWords(input1));
+
+        // Test case 2: input with multiple words
+        String input2 = "Hello world, how are you?";
```
chiaryan, line 38: Perhaps you can consider wrapping your tests 2-5 in separate test functions, so that you can run them independently from each other?

src/test/java/Martin/TaskListTest.java
```diff
@@ -0,0 +1,33 @@
...
+        // Create a new TaskList object
+        TaskList taskList = new TaskList(new ArrayList<Task>());
+
+        // Add some tasks to the task list
+        taskList.add(new Todo("Task 1"));
+        taskList.add(new Todo("Task 2"));
+        taskList.add(new Todo("Task 3"));
+
+        // Redirect the standard output to a ByteArrayOutputStream
+        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
```
chiaryan: Oo, didn't know you could redirect sysout like this, very nice way to test the print function.

src/main/java/Martin/Command.java
```diff
@@ -0,0 +1,93 @@
...
+     * @throws IOException If an I/O error occurs while accessing the storage.
+     */
+    public void handleCommand(ChatbotKeyword command, String remainingWords) throws IOException {
+        switch (command) {
+            case LIST:
+                taskList.printList();
+                break;
+            case MARK:
+                int doneIndex = Integer.parseInt(remainingWords);
+                taskList.markAsDone(doneIndex);
```
chiaryan: For methods that access the elements of the Tasklist, Is there any exception handler for the case where it tries to access an out-of bounds index?

chiaryan (commented):
Good job, couldn't find any issues w.r.t. coding standard. See my comments for possible improvements.

src/main/java/Martin/Command.java
tanyunchao: At this current point no. But definitely a possible extension of the project, since index out of bounds definitely happens. Thanks for the heads up

PR #274, opened 2024-01-30
[jeremiahlzz] iP

jeremiahlzz:
# a heading

## a bullet list
- 1
- 2
- 3

## a numbered list
1. 1
2. 2
3. 3

## a fenced code block (with syntax highlighting)
```java
    // Do Nothing
}
```

## a task list
- [ ] task a
- [x] task b
- [ ] task c

## an emoji
🐀 🐂 🐯 🐰 🐉 🐍 🐴 🐐 🐒 🐓 🐶 🐷 

## a hyperlink
[GitHub GFMD guide](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax)

## inline code
This is some inline code: `int i = 0;`.

## some text formatting: bold, italic, strikethrough etc.
**Lorem** _Ipsum_ <sup>small words</sup>


src/main/java/controller/Zero.java
```diff
@@ -0,0 +1,164 @@
...
+                    int idx = Parser.parseIndex(input.get("name"));
+                    Task t = storage.getTaskList().mark(idx);
+                    storage.saveTaskList();
+                    ui.showMarkDone(t);
+                } catch (NumberFormatException e) {
+                    ui.showIndexParseError();
+                } catch (IndexOutOfBoundsException e) {
+                    ui.showIndexOutOfBoundsError(storage.getTaskList());
+                }
+                break;
```
delishad21: Can consider abstracting the code for each case into new functions as your run code looks a little crowded. Maybe can also consider nesting the entire switch case in a try block and catching all the exceptions after the switch case has run? Since the exceptions from most of your cases are the same.

src/main/java/model/ToDo.java
```diff
@@ -0,0 +1,29 @@
+package model;
+public class ToDo implements Task {
+    private final String name;
+    private final boolean done;
+
+    ToDo(String name) {
```
delishad21: Missing access modifier here for constructor. Is the constructor meant to be package-private?

src/main/java/view/Ui.java
```diff
@@ -0,0 +1,97 @@
...
+    private String divider;
+
+    public Ui(Reader input, Writer output, String d) {
+        br = new BufferedReader(input);
+        pw = new PrintWriter(output) {
+            @Override
+            public PrintWriter printf(String format, Object... args) {
+                return super.printf(divider + "\n" + format + "\n" + divider + "\n", args);
+            }
+        };
```
delishad21: This is an interesting and useful way to set up bot responses!

src/test/java/controller/ParserTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+public class ParserTest {
+    @Test
+    public void parseInput_deadlineTaskInput_success() {
+        Hashtable<String, String> dict = Parser.parseInput("deadline testinput /by 21/2/23 1300");
+        assertAll("test",
+            () -> assertEquals("deadline", dict.get("cmd")),
+            () -> assertEquals("testinput", dict.get("name")),
+            () -> assertEquals("21/2/23 1300", dict.get("/by"))
+        );
```
delishad21, line 18: Very clean testing code. Easy to understand

delishad21 (commented):
Very neat code that is well thought out and organised into appropriate packages

src/main/java/controller/Zero.java
jeremiahlzz: Abstracting out each switch case as a function is a great idea! 😄 I will definitely consider it in future projects to improve readability.
Catching all exceptions after the switch case could potentially run into problems in the future, if I want a different UI response for each command. That being said its great idea for the current scope of the project 👍

src/main/java/model/ToDo.java
jeremiahlzz: Thanks for catching it!

PR #276, opened 2024-01-30
[CJerrong] iP

CJerrong:
## Haro

> “I love lists. Always have.” – Adam Savage ([source](https://medium.com/@oaklinejournal/25-quotes-from-successful-people-on-the-power-of-writing-to-do-lists-8a0b3729a34c))

Haro frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use
All you need to do is,

1. download it from [here](https://github.com/CJerrong/ip/releases).
2. double-click it.
3. add your tasks.
And it is **FREE**!

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)
---
If you are a Java programmer, you can use it to practice Java too. Here's the ``main`` method:
``` java
public class Main {
    }
}
```


.gitignore
```diff
@@ -15,3 +15,6 @@ bin/
 
 /text-ui-test/ACTUAL.TXT
 text-ui-test/EXPECTED-UNIX.TXT
+
+/data/
+data/saveList.txt
```
jeremiahlzz, line 20: `/data/` includes all files and directories within the `data` file.
Consider removing the redundant `data/saveList.txt`, or removing `/data/` to only exclude `saveList.txt` from commits.

src/main/java/haro/Parser.java
```diff
@@ -0,0 +1,214 @@
...
+        else if (instruction == Instruction.FIND){
+            commandArg = commandArg.trim();
+
+            if (commandArg.equals("")) {
+                throw new InvalidArgsException("Please input a valid search string\n");
+            }
+
+            FindCommand findCommand = new FindCommand(commandArg);
+            resultCommand = findCommand;
+        }
```
jeremiahlzz: The `switch case` and the `if-else` statements both accomplish the same thing, differentiate the instruction to be executed.
Consider using `Instruction.valueOf(instructWord.toUpperCase())` to convert the `String` input into an `Enum` type directly.

Should each instruction code also be abstracted out (as methods) to improve readability?

src/main/java/haro/Storage.java
```diff
@@ -0,0 +1,167 @@
...
+                        break;
+                    }
+                } else if (actionType == "deadline") {
+                    try {
+                        if (storageArgs.length < 4) {
+                            throw new Exception("Invalid save list");
+                        }
+
+                        int markedInt = Integer.valueOf(storageArgs[1].trim());
+                        boolean marked = markedInt == 0 ? false : true;
```
jeremiahlzz: According to the coding standards, ternary expressions require braces around the boolean expression. This improves readability. For example:
```java
alpha = (aLongBooleanExpression) ? beta : gamma;
alpha = (aLongBooleanExpression)
        ? beta
        : gamma;
```

src/main/java/haro/Storage.java
```diff
@@ -0,0 +1,167 @@
...
+                        break;
+                    }
+                } else if (actionType == "event") {
+                    try {
+                        if (storageArgs.length < 5) {
+                            throw new Exception("Invalid save list");
+                        }
+
+                        int markedInt = Integer.valueOf(storageArgs[1].trim());
+                        boolean marked = markedInt == 0 ? false : true;
```
jeremiahlzz: Ternary expression needs braces for the boolean expression.

src/main/java/haro/Storage.java
```diff
@@ -0,0 +1,167 @@
...
+                        resultTasks = new ArrayList<Task>();
+                        break;
+                    }
+                }
+            }
+
+        } catch (NoSuchElementException e) {
+
+        } catch (Exception e) {
+
```
jeremiahlzz: Empty catch blocks. Should a comment be included to declare that it was left empty on purpose? E.g. `// Do Nothing`

jeremiahlzz (commented):
Nice use of custom exceptions to handle erroneous user input!

PR #283, opened 2024-01-31
[joelgoh1] iP

joelgoh1:
# Blawg 
> ":D" - me 

Blawg 🐱 frees your mind of having to remember things you need to do. It's,
- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,

1. download it from [here](https://github.com/joelgoh1/ip).
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 😆 

And it is **FREE**!

Features:

- [x] Managing tasks
-  [ ] Managing deadlines (coming soon)
-  [ ] Reminders (coming soon)

```java
        new Blawg().run();
    }
```


src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,95 @@
...
+            FileWriter writer = new FileWriter(fileName, false);
+            writer.write(taskListWriter(data) + "\n");
+            writer.close();
+        } catch (IOException e) {
+            throw new FileError("Problem writing to file!");
+        }
+    }
+
+    public ArrayList<Task> read() throws FileError, FileAccessError {
+        ArrayList<String> content = new ArrayList<>();
```
ashleygoh1: contents (plural form)
since the array will store many contents

ashleygoh1 (commented):
LGTM! amazing code.

src/main/java/Blawg.java
```diff
@@ -0,0 +1,58 @@
+import commands.Command;
+import exceptions.BlawgException;
+import parser.Parser;
+import storage.Storage;
+import tasklist.TaskList;
+import ui.Ui;
```
Arixeyeion, line 8: I like that everything is split up into packages.

src/main/java/commands/DeleteCommand.java
```diff
@@ -0,0 +1,29 @@
...
+        this.index = index;
+    }
+
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws FileError {
+        Task removedTask = tasks.getTaskList().get(index - 1);
+        tasks.getTaskList().remove(index - 1);
+        String result = "Noted. I've removed this task:\n"
+                + removedTask
+                + String.format("Now you have %d tasks in the list\n", tasks.getTaskList().size());
```
Arixeyeion, line 29: For this portion, I would suggest starting the line with String.format. There is no need to + removedTask as that already uses the toString method. Instead doing String.format("Noted...\n%s...", removedTask.toString()...) would allow you to view the entire string as if you were reading a whole sentence or paragraph.

Arixeyeion (commented):
Overall, I like how everything is compartmentalized, this makes everything easier to understand as each class performs a "single" task. I also enjoyed the personalization of the messages the bot produced. Just some nits about the usage of String.format().

PR #293, opened 2024-01-31
whelan-low

whelan-low:
# **Greg the Assistant**

> "Your mind is for having ideas, not holding them" - David Allen

A chat-bot to free your brain from remembering your tasks! It can:
- Help you find your tasks easily!
- Store as many things as you need
- Supports many types of tasks!
- Extremely simple to use!

All you need to do is:
1. Download it from [here](github.com/whelan-low/ip)
2. double-click it
3. add your tasks
4. Let Greg handle the rest!

It is completely **FREE!!**

Features:
- [x] Task management
- [ ] Deadline management (Coming soon!)
- [ ] Event management (Coming soon!)

If you are a Java programmer, you can use it to practice Java too.
Here is the main method:
`     public static void main(String[] args) {
        Ui.startup();
    } `


src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,265 @@
...
+
+    /**
+     * Returns the task with the specified index from the list
+     *
+     * @param index the index of the list to be retrieved
+     * @return the task with that index
+     */
+    public Task get(int index) {
+        return tasks.get(index);
+    }
```
kinjalagarwal1810, line 46: Perpahs the name of the variable could be more specific like "getTask" so it is easily understandable

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,265 @@
...
+        return tasks.size();
+    }
+
+    /**
+     * Creates a string representation of the task list.
+     * If empty, returns an empty message.
+     *
+     * @return the string representation of the task list.
+     */
+    public String list() {
```
kinjalagarwal1810, line 63: You can consider naming your method here more specifically like "printList" so we know it is printing or returning a string.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,265 @@
...
+        }
+        return successful;
+    }
+
+    /**
+     * Announces the addition of a task
+     *
+     * @param task the task being added
+     */
+    public void announceAddition(Task task) {
```
kinjalagarwal1810, line 280: Perhaps instead of annouce you can use print since it is more commonly used when we return a string

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,77 @@
...
+        fillerLine();
+        System.out.println("    Hello! I am Greg.\n    What can I do for you?");
+        fillerLine();
+    }
+
+    /**
+     * Runs the shutdown sequence.
+     * Saves then says bye.
+     */
+    public void bye() {
```
kinjalagarwal1810: The name of method would be better if it was a verb so maybe use "printBye" or another verb.

kinjalagarwal1810 (commented):
Overall looks very well done! I've just added a couple of minor comments.

PR #295, opened 2024-01-31
[willowisp01] iP

willowisp01:
# Luke

> Like Duke, but starting with L instead. -- (Myself)

Luke is like Duke but better. That's because you can
- add different types of events
- type commands, like, really fast
- form a unique bond with Luke

You can download it [here](https://github.com/willowisp01/ip/releases/tag/A-Jar) for the price of **zero dollars and zero cents**!

Features: 

- [x] Managing Tasks 
- [x] Managing Deadlines
- [x] Tsundere Personality 💢💢💢 👉👈
- [ ] Other Personalities (WIP)

The full source code:
```
   while (true) {
      respondToInput(input)
   }
}
```


src/main/java/luke/TaskList.java
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * Looks for a specified task in the tasklist using the given keyword.
+     * Prints them out, as well as their position in the list (index + 1), if any are found.
+     *
+     * @param keyword The keyword to search for a task.
+     */
+    public void findTask(String keyword) {
+        int num = 1;
+        boolean found = false;
```
Alteqa: Should be isFound or hasBeenFound instead of simply just found.

src/main/java/luke/UI.java
```diff
@@ -0,0 +1,26 @@
+package luke;
+
+public class UI {
+    private static String logo = "  _____             __             \n"
```
Alteqa: constants should be all capitalised such as LOGO instead of just logo.

src/main/java/luke/Luke.java
```diff
@@ -0,0 +1,78 @@
...
+import java.nio.file.FileAlreadyExistsException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Scanner;
+
+public class Luke {
+
+    //Logo created using https://patorjk.com/software/taag/#p=display&f=Varsity&t=Luke
+    private static String logo = "  _____             __             \n"
```
Alteqa: Constants should be capitalised.

src/main/java/luke/TaskList.java
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * Adds task to tasklist.
+     *
+     * @param task The task to be added.
+     */
+    public void addTask(Task task) {
+        if (task != null) {
+            tasks.add(task);
+            System.out.println("I helped you add task '" + task.fullStatus() + "'. But do it yourself next time! Hmmph!" + "\n");
```
Alteqa: code should be wrapped/split by the + instead of the string itself.

src/main/java/luke/Luke.java
```diff
@@ -0,0 +1,78 @@
...
+            + "   | |   _[  | | |  | '' < / /__\\\\ \n"
+            + "  _| |__/ || \\_/ |, | |`\\ \\| \\__., \n"
+            + " |________|'.__.'_/[__|  \\_]'.__.' ";
+
+    public static void main(String[] args) throws LukeException {
+        Storage storage;
+        UI ui = new UI();
+        ui.greet();
+
+        String wd = System.getProperty("user.dir");
```
Alteqa: variable names should be descriptive.

Alteqa (commented):
Overall your code was excellent and there were very few errors. Even with errors that I found they were very minor and did not affect my understanding of your programme at all.

src/main/java/luke/Deadline.java
```diff
@@ -0,0 +1,41 @@
+package luke;
+
+import java.time.LocalDate;
+
+public class Deadline extends Task {
```
Austintjh19, line 5: Maybe "DeadlineTask" as a class name might be more relevant and descriptive compared to simply "Deadline." Might have a Deadline task in the future. Applicable to Event & Todo as well. :)

src/main/java/luke/Deadline.java
```diff
@@ -0,0 +1,41 @@
...
+    private LocalDate by;
```
Austintjh19: byDate may be more applicable as a variable name. Given that you are storing it as a LocalDate. Future programmers will know it doesn't store for instance Tuesday or 3 pm.  Applicable to Event as well.

src/main/java/luke/Luke.java
Austintjh19: The process of initialising the storage file may be more suitable to be done by the Storage class.

src/main/java/luke/Luke.java
```diff
@@ -0,0 +1,78 @@
...
+        } catch (FileAlreadyExistsException e) {
+            //file already exists.
+        } catch (IOException e) {
+            System.out.println("Failed to create file");
+            return;
+        }
+
+        //load the file if there is save data (reference: https://www.baeldung.com/java-serialization)
+        //otherwise, create a new History object.
+        File historyFile = new File(String.valueOf(historyPath));
```
Austintjh19: The process of loading the storage file may be more suitable to be done by the Storage class.

src/main/java/luke/Parser.java
```diff
@@ -0,0 +1,142 @@
...
+
+    /**
+     * Parses a command given a string input. Returns false if there is a next command, else true.
+     *
+     * @param input String input as given by the user.
+     * @return False if there is a next command, else true.
+     */
+    public boolean parseCommand(String input) {
+        String trimmedLowercase = input.trim().toLowerCase();
+        if (trimmedLowercase.equals("bye")) {
```
Austintjh19: Maybe consider switching to a case statement rather than relying on if-else constructs, it could potentially enhance code readability.

src/main/java/luke/Storage.java
```diff
@@ -0,0 +1,49 @@
+package luke;
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.ObjectOutputStream;
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.io.IOException;
+
+public class Storage implements Serializable {
```
Austintjh19, line 10: Considering future iterations of Luke, it may be beneficial to create a separate class, such as TxtStorage, that inherits from this file. This approach allows for potential expansion beyond storing data solely in text files in future versions of the application. Just a suggestion.

src/main/java/luke/Storage.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private ArrayList<Task> history;
```
Austintjh19, line 12: Maybe a more appropriate name ? e.g. TasksHistory

src/main/java/luke/Task.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    public boolean isDone() {
+        return isDone;
+    }
+
+    /**
+     * Marks this task as complete.
+     */
+    public void complete() {
```
Austintjh19, line 30: Maybe setIsDone is more appropriate ?

src/main/java/luke/Deadline.java
```diff
@@ -0,0 +1,41 @@
...
+    public Deadline(String name, LocalDate by) {
+        super(name);
+        this.by = by;
+    }
+
+    /**
+     * @return The status of task. ([taskType][isDone] taskname by)
+     */
+    @Override
+    public String fullStatus() { //TODO: add type of task later
```
Austintjh19: I think using a verb as part of the function name is more appropriate. E.g. getFullStatus ?

src/main/java/luke/TaskList.java
```diff
@@ -0,0 +1,112 @@
...
+     *
+     * @param index Index of task to be marked as done.
+     */
+    public void markTask(int index) {
+        try {
+            tasks.get(index).complete();
+        } catch (IndexOutOfBoundsException e) {
+            System.out.println("Jeez, you really ought to give me a number I can work with... got that?");
+            System.out.println("[Item index exceeds history count]\n");
+            return;
```
Austintjh19: Maybe return a boolean. True/ false representing whether the task was successfully executed. Instead of having an empty return.

Austintjh19 (commented):
Good work. really well written code. Just a few suggestion here and there from me.

PR #301, opened 2024-01-31
[youdonnnn] iP

youdonnnn:
### Introducing the YoudonBot!
 > Sign up for a free trial now! [link here](https://en.wikipedia.org/wiki/Cat)

YoudonBot allows you to focus on the more important things in life. We are: 
* text-based
* easy to learn
* ~~FAST~~ _SUPER_ FAST to use

All you need to do ___right now___ is:
1. download it [here](https://en.wikipedia.org/wiki/Dog).
2. double-click the __.exe__ file.
3. add in your tasks and events.
4. and let it do the rest of the _magic_ for you &#x1f604;

And the best part is that it is __FREE__!

Features include:
- [x] Managing tasks
- [x] Managing deadlines (new release!)
- [ ] Reminders (coming soon)
***
```Java
public class Main {
    }
}

### CODE REVIEW

File: src/main/java/Deadline.java [hunk 1]
```diff
@@ -0,0 +1,17 @@
+public class Deadline extends Task {
```
jieqiboh: Perhaps a Javadoc comment describing the class would be useful 😃 
I noticed the same issue in other places too

src/main/java/Task.java
```diff
@@ -0,0 +1,17 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
```
jieqiboh: I like the way you named this boolean variable 😃

src/main/java/Event.java
```diff
@@ -0,0 +1,19 @@
...
+        this.end = end;
+    }
+
+    public String getTypeIcon() {
+        return "E";
+    }
+
+    @Override
+    public String toString() {
+        return "[" + this.getTypeIcon() + "][" + this.getStatusIcon() + "] " + this.description + "(from: " + this.start + "to:" + this.end + ")";
```
jieqiboh:
```suggestion
        return "[" + this.getTypeIcon() + "][" + this.getStatusIcon() + "] " 
        + this.description + "(from: " + this.start + "to:" + this.end + ")";
```
I think splitting the code into a newline would be useful 😃

src/main/java/Event.java
```diff
@@ -0,0 +1,19 @@
+public class Event extends Task {
+    protected String start;
+    protected String end;
+
+    public Event(String description,  String start, String end) {
```
jieqiboh:
```suggestion
    public Event(String description, String start, String end) {
```
I think there was a minor typo where you used 2 spaces to separate non-whitespace characters instead of 1

src/main/java/Deadline.java
```diff
@@ -0,0 +1,17 @@
...
+        this.deadline = deadline;
+    }
+
+    public String getTypeIcon() {
+        return "D";
+    }
+
+    @Override
+    public String toString() {
+        return "[" + this.getTypeIcon() + "][" + this.getStatusIcon() + "] " + this.description + "(by: " + this.deadline + ")";
```
jieqiboh:
```suggestion
        return "[" + this.getTypeIcon() + "][" + this.getStatusIcon() + "] " + this.description + "(by: " 
        + this.deadline + ")";
```
Perhaps splitting the code into a newline would be useful 😃

jieqiboh (commented):
LGTM 😄 
Although there were a few minor coding standard violations, I found your code easy to read for the most part, and the current issues should be fairly easy to fix.

src/main/java/youdon/Deadline.java
```diff
@@ -0,0 +1,24 @@
...
+        return "D";
+    }
+
+    @Override
+    public String toString() {
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("MMM dd yyyy HH:mm");
+        String formattedDateTime = this.deadline.format(formatter);
+        return "[" + this.getTypeIcon() + "][" + this.getStatusIcon() + "] " + this.description + "(by: " + formattedDateTime + ")";
+    }
+}
```
chiralcentre: Should String.format be used instead of concatenation to make the code cleaner? I noticed this nit for the other classes as well.

src/main/java/youdon/Parser.java
chiralcentre, line 1: Should a switch statement be used to clearly highlight all the different logic branches?

src/main/java/youdon/Storage.java
chiralcentre, line 1: Should the return type of loadData() be TaskList instead of ArrayList<Task>?

src/main/java/youdon/Youdon.java
chiralcentre, line 1: Should the business logic in parser.parse() be in the main class here instead? I was thinking it doesn't make much sense for the parser to handle the business logic for the execution of the application; instead the parser should only be responsible for extracting the command for each line of input.

src/main/java/youdon/validCommands.java
chiralcentre, line 1: Would it be good to abstract out the command types into separate classes as per the requirements of the A-MoreOOP increment?

chiralcentre (commented):
LGTM, but I have highlighted some issues, which I hope will be of use to you.

PR #304, opened 2024-01-31
[wongkj12] iP

wongkj12:
# Duke

> "Hey man, i'm MR. WONG. How may I help you today?" - Duke

**Duke** is a task manager chatbot that can help you to
- Remember todo lists
- Set deadlines
- Keep track of events
- and many more!

All you need to do is,
1. Download it from [here](https://github.com/wongkj12/ip/releases)
2. Run it
3. Add your tasks using _text-based_ commands
4. Let it manage your tasks! 🗿  

Current features:
- [x] Managing todos
- [x] Deadlines
- [x] Events
- [ ] Reminders (maybe coming soon)

Here is the current `main` method:
```java
public class Duke {
    public static void Duke(String[] args) {
        // ...
        }
}
```


src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,84 @@
+package duke;
+
+import duke.command.*;
```
Wongqingyee, line 3: Avoid using * wildcard.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,144 @@
...
+import duke.task.Event;
+import duke.task.Task;
+import duke.task.Todo;
+
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.util.List;
+import java.util.ArrayList;
+import java.util.Scanner;
+import java.io.File;
```
Wongqingyee, line 13: Ordering of the import statements should be consistent. 
It should be arranging all the java.io. first, then followed by java.util.

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * TaskList class to represent the collection of tasks being maintained.
+ */
+public class TaskList {
+    private List<Task> list;
```
Wongqingyee, line 19: Plural form should be used on names representing a collection of objects.

src/test/java/duke/ParserTest.java
```diff
@@ -0,0 +1,28 @@
+package duke;
+
+import org.junit.jupiter.api.Test;
```
Wongqingyee, line 3: The import statements should be in the following order:
import static ...
<blank line>
import ...

src/test/java/duke/task/DeadlineTest.java
```diff
@@ -0,0 +1,19 @@
+package duke.task;
+
+import org.junit.jupiter.api.Test;
```
Wongqingyee, line 3: Same issue for the ordering of the import statements.

src/main/java/duke/Parser.java
dedsecrattle, line 3: Can we avoid widlcard imports in order to adhere to the coding style ?

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,19 @@
...
+import duke.Ui;
+import duke.task.Task;
+
+public class DeleteCommand extends Command {
+    private int index;
+    public DeleteCommand(int index) {
+        this.index = index;
+    }
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
```
dedsecrattle: Can we add javadoc for the method `execute`?

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,52 @@
...
+        this.from = from;
+        this.to = to;
+
+        // Check if they are in dateTime format
+        // If so, update the string
+        try {
+            fromDate = LocalDate.parse(from);
+            this.from = fromDate.format(DateTimeFormatter.ofPattern("MMM d yyyy"));
+
+        } catch (DateTimeParseException e) {}
```
dedsecrattle, line 38: Can we try to convey the user about this `DateTimeParseException` instead of just catching it ?

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,12 @@
...
+
+import duke.Storage;
+import duke.TaskList;
+
+public abstract class Command {
+    protected boolean isExit = false;
+    public boolean isExit() {
+        return this.isExit;
+    }
+    public abstract String execute(TaskList tasks, Storage storage);
```
jovantanyk, line 14: Could use higher level of abstraction here. Maybe have another class to handle UI related methods instead of returning String directly as there might be commands that dont require any output.

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,40 @@
...
+    public Deadline(String name, String by) {
+        super(name);
+        this.by = by;
+
+        // Check if they are in dateTime format
+        try {
+            byDate = LocalDate.parse(by);
+            this.by = byDate.format(DateTimeFormatter.ofPattern("MMM d yyyy"));
+        } catch (DateTimeParseException e) {}
+    }
```
jovantanyk: Try and avoid having too much business logic in class constructors. The input should already be parsed and processed by the Parser.java class instead of letting Deadline.java handle it

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,52 @@
+package duke.task;
+
+/**
+ * Class representing a generic Task
+ */
+public class Task {
```
jovantanyk, line 6: Since Task is never instantiated directly, you should make it an abstract class and allow subclasses to have their own implementation of the methods instead of overriding every method when Task is never instantiated to begin with

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,55 @@
...
+public class Storage {
+    private File file;
+    private String filepath;
+
+    /**
+     * Creates an instance of Storage.
+     * @param filepath
+     */
+    public Storage(String filepath) {
+        this.filepath = filepath;
```
jovantanyk, line 20: Since filepath will never change across the lifecycle of a program, it might be better to make Storage a static class instead, so that you won't have to pass an instance of Storage around every time you want to load or save a file

jovantanyk (commented):
Overall Good display of OOP knowledge and consistent method and class names.  Could use higher level of abstraction like creating a Ui class to handle output instead of having every Command deal with it directly as it creates a level of inconsistency.

PR #322, opened 2024-01-31
[camille-readbean] iP

camille-readbean:
# Leto 👑

> “Once you have explored a fear, it becomes less terrifying. Part of courage comes from extending our knowledge.” -Duke Leto Atreides [source](https://www.redbubble.com/life/dune-quotes/)    

Duke Leto helps you with keeping track of todos, deadlines and events. It's,

- text-based
- easy to learn
- ~~FAST~~ _SUPER FAST_ to use
  
All you need to do is,

1.    download it from [here](https://nus-cs2103-ay2324s2.github.io/website/schedule/week4/project.html).
2.    double-click it.
3.    add your tasks.
4.    let it manage your tasks for you 😉

And it is **FREE**!

Features:

- [X]    Managing tasks
- [X]    Managing deadlines 
- [X]    Managing events
- [X]    Extensible way to add commands 🔧 
- [X]    Save 💾  and load from csv 
- [ ]     Reminders (coming soon)

```java
public class Main {
        letoLogo();
        TaskList.initFromFile();
        Parser.readCommandAndExecute();
    }
}
```


Leto-Tasks.csv
ZHANGTIANYAO1, line 1: Storing data with CSV is a very good idea, learn from you. 👍

src/main/java/leto/parser/Parser.java
```diff
@@ -0,0 +1,35 @@
...
+import static leto.ui.Ui.shortSay;
+
+public class Parser {
+    private static Scanner sc = new Scanner(System.in);
+
+    public Parser() {
+    }
+
+    public static void readCommandAndExecute() {
+        boolean bye = false;
```
ZHANGTIANYAO1: Could be `isBye` or `isExit`

src/main/java/leto/tasklist/BadTaskIndexException.java
```diff
@@ -0,0 +1,10 @@
+package leto.tasklist;
+
+public class BadTaskIndexException extends InvalidTaskException {
+    public BadTaskIndexException(int CurrentIndex) {
+        super(
+            "Bad task index number, should be a integer more than 0 and less than: " +
+            (CurrentIndex+1) +
+            " see /list for all the tasks");
```
ZHANGTIANYAO1: Might be better to move `+` in front.
I noticed the same issue in several other places too

src/main/java/leto/tasklist/Deadline.java
```diff
@@ -0,0 +1,71 @@
...
+public class Deadline extends Task {
+    private LocalDate deadline;
+
+    private Deadline(Boolean completed, String TaskString, LocalDate deadline) {
+        // Call to 'super()' must be first statement in constructor body
+        super(completed, TaskString);
+        this.deadline = deadline;
+    }
+
+    public static Deadline DeadlineFromCMD(String input) throws InvalidTaskException {
```
ZHANGTIANYAO1: Might be better to rename as  `deadlineFromCMD`
I noticed the same issue in several other places too

src/main/java/leto/tasklist/Event.java
```diff
@@ -0,0 +1,75 @@
...
+            return Math.abs(daysToStart) + " days to the event";
+        } else if (daysToEnd < 0) {
+            return "ended " + Math.abs(daysToEnd) + " days ago";
+        } else {
+            return daysToEnd + " days remaining";
+        }
+    }
+
+    public static Event EventFromCMD(String input) throws InvalidTaskException {
+        String regex = "(?i)event ([^,]+) /from (\\d{4}-\\d{2}-\\d{2}) /to (\\d{4}-\\d{2}-\\d{2})";
```
ZHANGTIANYAO1: Wow 👍

src/main/java/leto/tasklist/TaskList.java
```diff
@@ -0,0 +1,141 @@
...
+            switch (typeOfTask.toLowerCase()) {
+                case "event":
+                    t = Event.EventFromCMD(inputs);
+                    break;
+                case "deadline":
+                    t = Deadline.DeadlineFromCMD(inputs);
+                    break;
+                case "todo":
+                    t = Todo.TodoFromCMD(inputs);
+                    break;
```
ZHANGTIANYAO1: I think there should be no indentation for case clauses

ZHANGTIANYAO1 (commented):
LGTM. 👍
Just some nits to fix.
I learned a lot.

Leto-Tasks.csv
camille-readbean: Thank you for your inputs. 😃    
  
I must point out that due to the use of CSV, we limit our inputs to not have commas.  
A better solution would be changing our delimiter to other characters like `|` that is usually not input by the user.  
There is another variant called `tsv` which is just using tab characters `\t` to delimit the entries.  
  
This was quite a hacky solution as the standard java library do not contain any out of the box libraries unlike Python 🐍

src/main/java/leto/parser/Parser.java
camille-readbean: Will fix upstream! Thanks for the lookout

src/main/java/leto/tasklist/BadTaskIndexException.java
camille-readbean: 👷  will fix!

src/main/java/leto/tasklist/Deadline.java
camille-readbean: Thanks for catching that!

Leto-Tasks.csv
```diff
@@ -0,0 +1,7 @@
+D,N,20,2021-03-02,,
```
owenong1: Not sure if you want to commit the data file as well. 
Maybe could specify it as a default pre-loaded list.

src/main/java/me/ruibin/leto/tasklist/InvalidTaskException.java
```diff
@@ -0,0 +1,15 @@
+package me.ruibin.leto.tasklist;
+
+import me.ruibin.leto.ui.Ui;
+
+/** Exception for invalid input for a Task */
+public class InvalidTaskException extends Exception {
```
owenong1, line 6: It seems like almost all the exceptions extend InvalidTaskException; even ones like InvalidCmdException.
To me its not intuitive that an invalid command constitutes an invalid task.
Perhaps the naming/inheritance could be modified to make it more intuitive.
Otherwise, I think it makes sense!

src/main/java/me/ruibin/leto/tasklist/TaskList.java
```diff
@@ -0,0 +1,172 @@
...
+import java.util.List;
+
+import me.ruibin.leto.storage.Storage;
+import me.ruibin.leto.ui.Ui;
+
+/** Class representing the list of tasks in Leto. */
+public class TaskList {
+    private static final ArrayList<Task> list = new ArrayList<>(100);
+
+    public TaskList() {}
```
owenong1: A bit minor but I think Java will fill in this constructor by default so you may not need this line

src/main/java/me/ruibin/leto/tasklist/TaskList.java
```diff
@@ -0,0 +1,172 @@
...
+
+    /** Static method to initialise the task list from csv file. */
+    public static void initFromFile() {
```
owenong1: Since you used csv to refer to the file in other methods, maybe initFromCsv would be more informative

src/main/java/me/ruibin/leto/tasklist/TaskList.java
```diff
@@ -0,0 +1,172 @@
...
+    /**
+     * Provides a way to add a task to task list.
+     * For example in Storage class readEntry.
+     * */
+    public static void addTaskToList(Task t) {
+        list.add(t);
+    }
+
+    /** In charge of reading inputs and adding the corresponding task to the list */
+    public static void addToList(String inputs) {
```
owenong1: I think you can use overloading here! Since the functionality of adding to list is the same, but the argument types are different it may not be necessary to create another named function.

owenong1 (commented):
GJ Rui Bin!! I think your code is pretty solid overall in terms of code practices and overall logic. Only thing I can think of is to adjust some minor naming/structural aspects, but LGTM.

ATB FOR THE REST OF THE SEM!!

Leto-Tasks.csv
camille-readbean: I think this was committed accidentally should be in the git ignore in the future I think!

src/main/java/me/ruibin/leto/tasklist/TaskList.java
camille-readbean: Oh right I forgot I could have done that!

PR #327, opened 2024-01-31
[chiralcentre] iP

chiralcentre:
Felix helps you to organise and keep track of your upcoming tasks. Felix is:

- text-based
- intuitive
- ~~FAST~~ *SUPER* FAST to use

All you need to do is,

1. download it from [here](https://github.com/chiralcentre/ip).
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 😄 

And it is FREE!

Features:

- [ ] Managing upcoming tasks to be completed without a fixed deadline
- [ ] Managing upcoming tasks to be completed with a fixed deadline
- [ ] Managing upcoming events with a fixed start and end time
- [x] Managing reminders

If you are a Java programmer, you can use it to practice Java too. Here's the `main` method in `Felix.java`:
```java
public static void main(String[] args) throws IOException, FelixException {
    Felix felix = new Felix();
    felix.run();
}
```


src/main/java/felix/utils/Parser.java
```diff
@@ -0,0 +1,89 @@
...
+                throw new FelixException("Command does not follow this format: deadline {description} by {end_datetime}");
+            } catch (DateTimeParseException e) {
+                throw new FelixException("datetime for deadline is not in the format \"yyyy-MM-dd HHmm\"");
+            }
+        } else {
+            try {
+                String[] remainder = words[1].split(" /from | /to ");
+                task = new Event(remainder[0], remainder[1], remainder[2]);
+            } catch (IndexOutOfBoundsException e) {
+                throw new FelixException("Command does not follow this format: event {description} /from {start_datetime} /to {end_datetime}");
```
yisiox: Should this line be wrapped? I noticed the same issue in other places too.

src/main/java/felix/utils/Storage.java
```diff
@@ -0,0 +1,81 @@
...
+        } else {
+            System.out.println("file already exists");
+        }
+    }
+
+    /**
+     * Generates a list of tasks from the storage file associated with the created Storage instance.
+     *
+     * @return list of tasks.
+     * @throws FelixException If file is not found, there is error reading from the file, or datetime is not in correct format.
```
yisiox: Suggestion for formatting long comment:
```
@throws FelixException  If file is not found, there is error reading from the file, 
                        or datetime is not in correct format.
```

src/main/java/felix/command/MarkCommand.java
```diff
@@ -0,0 +1,26 @@
+package felix.command;
+
+import felix.exception.FelixException;
+import felix.task.Task;
+import felix.utils.TaskList;
+import felix.utils.Ui;
+import felix.utils.Storage;
```
yisiox: Nit: perhaps add a line break here, and in some other places.

src/main/java/felix/task/Task.java
```diff
@@ -0,0 +1,70 @@
...
+        } else if (words[0].equals("D")) {
+            task = new Deadline(words[2], words[3]);
+        } else {
+            task = new Event(words[2], words[3], words[4]);
+        }
+        if (words[1].equals("X")) {
+            task.markAsDone();
+        }
+        return task;
+    }
```
yisiox: I like how the `Task` class handles creation of the task objects so task types are hidden from `Storage`! Overlooked this myself. A nit may be a switch statement will make this look cleaner.

src/main/java/felix/task/Task.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Override
+    public String toString() {
+        return String.format("[%s] %s",this.getStatusIcon(),this.description);
```
yisiox: Nit: spaces can be added after commas. I noticed the same issue in other places too.

src/main/java/felix/utils/Parser.java
```diff
@@ -0,0 +1,89 @@
...
+            } catch (DateTimeParseException e) {
+                throw new FelixException("datetime not in the format \"yyyy-MM-dd HHmm\"");
+            }
+        }
```
yisiox: Nit: is a switch statement cleaner here?

src/main/java/felix/utils/TaskList.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /**
+     * Returns the String representation of all tasks within taskList to be written to a file.
+     */
+    public String getFileRepresentation() {
+        StringBuilder output = new StringBuilder();
+        for (int i = 0; i < this.tasks.size(); i++) {
+            output.append(this.tasks.get(i).toFileString());
+            if (i < this.tasks.size() - 1) output.append("\n");
```
yisiox: Single statement conditionals should still be wrapped by curly brackets. I noticed the same issue in other places as well.

src/main/java/felix/utils/Ui.java
```diff
@@ -0,0 +1,85 @@
...
+
+public class Ui {
+    private static final String BOT_NAME = "Felix";
+    private static final String LOGO = "___________    .__  .__        \n"
+            + "\\_   _____/___ |  | |__|__  ___\n"
+            + " |    __)/ __ \\|  | |  \\  \\/  / \n"
+            + " |     \\\\  ___/|  |_|  |>    <  \n"
+            + " \\___  / \\___  >____/__/__/\\_ \\ \n"
+            + "     \\/      \\/              \\/ \n";
+    private static final String LINE_SEPARATOR = "__________________________"
```
yisiox: Suggestion to make `LINE_SEPARATOR` size explicit: `LINE_SEPARATOR = "_".repeat(60);`

yisiox (commented):
LGTM. Just some nits to fix.

src/main/java/felix/task/Task.java
yisiox: After some thought, I think this can be improved further by moving this to a separate `TaskFactory` class so as to not have `Task` "know" its subclasses.

src/main/java/felix/utils/Parser.java
chiralcentre: Thank you for the feedback! Just to clarify, what do you mean by wrapping the line?

src/main/java/felix/command/MarkCommand.java
chiralcentre: Thank you for the feedback! Just to clarify, do you mean adding a line break between every import statement?

src/main/java/felix/task/Task.java
chiralcentre: Thank you for the feedback! Just to clarify, what is the purpose of the TaskFactory class?

src/main/java/felix/task/Task.java
chiralcentre: Thank you for the feedback! I will fix this.

src/main/java/felix/utils/TaskList.java
chiralcentre: Thank you for the feedback! I will fix this.

src/main/java/felix/utils/Ui.java
chiralcentre: Thank you for the feedback! I will fix this.

src/main/java/felix/utils/Parser.java
yisiox: Hi, the course coding standard states that lines should not pass the soft limit of 110 lines, and a hard limit of 120 lines. The oracle coding standard states 80 lines. This line is really long so it could be changed to look (for e.g.) like
```java
throw new FelixException("Command does not follow this format: event {description} "
        + "/from {start_datetime} /to {end_datetime}");
```

src/main/java/felix/task/Task.java
yisiox: Instead of having a `createTask` method inside the `Task` base class itself, which somewhat violates OOP since `Task` "knows" its subclasses, I made another class `TaskFactory` which is responsible for returning the `Task` of the relevant runtime type.

src/main/java/felix/utils/Parser.java
```diff
@@ -0,0 +1,101 @@
...
+ * Class representing a parser that processes user commands
+ */
+public class Parser {
+    /**
+     * Parses a line of user input and returns the corresponding Command.
+     *
+     * @return Command.
+     * @throws FelixException If command is unrecognised or command formatting is wrong.
+     */
+    public Command getCommand(String line) throws FelixException {
```
alfaloo, line 30: Do you think its better to split this long method (more than 30 lines) into smaller more readable methods?

PR #335, opened 2024-01-31
[PallonCX] iP

PallonCX:
# Killua

> "Just Let Me Handle This." - Killua Zoldyck ([source](https://www.cbr.com/hxh-best-killua-quotes/#quot-if-you-so-much-as-touch-them-i-39-ll-kill-you-quot))

Killua frees your mind of having to remember things you need to do. It's,
- Text-based
- Easy to learn
- FAST SUPER FAST to use

There are three type of tasks Killua can manage for now:
1. **Todo** task - Support task description.
2. **Deadline** task - Support task description, due date, and due time (time is optional).
3. **Event** task - Support task description, start date and time, end date and time (time is optional).

Killua can help you to store the tasks in your local device's memory. The only thing you need to do is to modify the file path in the code below (the code can be found in `Duke.java`):
```java
    new Duke("data/tasks.txt").run();
}
```

## Future RoadMap :sunglasses:

Features below may be implemented soon:
- [ ] Better Graphical User Interface (GUI).
- [ ] Higher tolerance to command given, and more acceptable commands.
- [ ] Applying Natural Language Processing (NLP) to improve user interaction experience.


data/tasks.txt
tankh99, line 1: Try to avoid committing storage files

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,220 @@
...
+
+            // Folder exist but not folder
+            if (i != splitedFile.length - 1 && !Files.isDirectory(path)) {
+                throw new DukeException();
+            }
+
+            // File not exist
+            if (i == splitedFile.length - 1 && !Files.exists(path)) {
+                try {
+                    Files.createFile(path);
```
tankh99, line 80: Why did you create a for loop just to check if each directory exists? I think that System.getProperty("user.dir") already ensures that the path it returns all exists

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,220 @@
...
+                for (String task: tasks) {
+                    String[] splited = task.split("`");
+                    switch (splited[0]) {
+                    case "duke.task.Deadline": {
+                        String content = splited[1];
+                        boolean isDone = splited[2].equals("Y");
+                        String deadline = splited[3];
+                        String[] splitedDateTime = deadline.split(" ");
+
+                        // Date and time
```
tankh99, line 113: nit: Perhaps give a more descriptive comment

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,220 @@
...
+                default: {
+                    textToAdd.append(t.getClass().getName() + "`" + t.getFields()[0] + "`" + t.getFields()[1] + "/end");
+                }
+            }
+        }
+
+        try {
+            FileWriter fw = new FileWriter(filePath.toString());
+            fw.write(textToAdd.toString());
+            fw.close();
```
tankh99, line 210: I think you should close fw in a finally block. Because this would not close in case an exception is thrown

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,348 @@
...
+
+/**
+ * The `Parser` class represents a tool to make sense of user input.
+ * It provides methods to verify and understand user input, then create respective commands.
+ */
+public class Parser {
+
+    public enum TaskType {
+        BYE, LIST, MARK, UNMARK, TODO, DEADLINE, EVENT, DELETE, FIND, UNKNOWN
+    }
```
tankh99: I think it could be good if this was extracted out to a class of its own

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,348 @@
...
+     * @return A command object to be executed.
+     * @throws IllegalArgumentException If any syntax error exists.
+     */
+    public static Command handleTaskBye(String[] splitedTask) throws DukeException {
+        // Incorrect command syntax handler
+        if (splitedTask.length > 1) {
+            throw new DukeException("Syntax of bye: bye");
+        }
+        return new ExitCommand();
+    }
```
tankh99: Although you're following naming convention, perhaps you could make the function more intuitive. E.g. handleTaskExit since "bye" could potentially be replaced with any other word, but the action would remain the same "exit application"

src/main/java/duke/command/Command.java
tankh99, line 1: I like abstraction and I like how you implemented it in this class

src/main/java/duke/command/Command.java
```diff
@@ -0,0 +1,30 @@
...
+     * @param tasks Existing tasks.
+     * @param ui The Ui of the program.
+     * @param storage The storage of the program.
+     * @throws DukeException For any error.
+     */
+    public abstract void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException;
+
+    /**
+     * Returns True if and only if the command is `ExitCommand`.
+
```
tankh99: nit: missing a *

src/main/java/duke/command/DeleteCommand.java
```diff
@@ -0,0 +1,59 @@
...
+        ui.delete(tasks.getItem(index), tasks);
+        tasks.delete(index);
+        storage.saveChanges(tasks);
+    }
+
+    /**
+     * Returns False.
+
+     * @return False.
+     */
```
tankh99: Could use a more descriptive description. E.g., what does isExit mean and what is it used for? It's not immediately apparent

nit: also missing another *

src/main/java/duke/task/Deadline.java
```diff
@@ -0,0 +1,132 @@
...
+    /**
+     * Returns the formatted string containing done status and description of the task.
+     *
+     * @return The formatted string containing done status and description of the task.
+     */
+    @Override
+    public String getDescriptionStatus() {
+        return "[D]" + (isDone ? "[X] " : "[ ] ") + this.description + " (by: "
+                + this.getDeadlineForDisplay() + ")";
+    }
```
tankh99, line 117: If you're overriding, you could actually create a toString() method in Task and then use that toString() method and modify it inside Deadline.

src/test/java/duke/task/DeadlineTest.java
```diff
@@ -0,0 +1,20 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class DeadlineTest {
+    @Test
+    public void getDeadlineForDisplayTest(){
```
tankh99: Does not follow standard test naming convention. Correct me if I'm wrong if this isn't the standard convention

`featureName_situation_expectedResult`. 

E.g. `getDeadline_validFormat_returnsFormattedDate`

src/main/java/duke/command/AddCommand.java
```diff
@@ -0,0 +1,172 @@
...
+
+    /**
+     * Returns False.
+
+     * @return False.
+     */
+    @Override
+    public boolean isExit() {
+        return false;
+    }
```
tankh99: if the superclass method already has a javadoc comment, no need to write comment for overridden method

tankh99 (commented):
I've reviewed your PR briefly, found just a bunch of nits. Mainly confused about your storage code and why you decided to do the things you did there

I also found a few coding standard violations but not a whole lot. You added comments to almost if not every function and class, good job! Although, I'm of the opinion that some comments can be left out if they're not particularly useful. And also, some comments that you wrote were not particularly descriptive either, like "isExit"

src/main/java/duke/DukeException.java
```diff
@@ -0,0 +1,22 @@
+package duke;
+
+/**
+ * The `DukeException` class represents any exception related to the Duke program.
+ */
+public class DukeException extends Exception{
```
simkianseng: Perhaps this was missed, the `Exception{` could have a space between the word `Exception` and the brace `{`.

> White space within a statement &ndash; se-education.org

Can refer to the documentation [here](https://se-education.org/guides/conventions/java/index.html).

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,348 @@
...
+        if (splitedTask.length == 1) {
+            throw new DukeException("Syntax of deadline: deadline {task description} /by ({yyyy-MM-dd HH:mm} or {yyyy-MM-dd})\n"
+                    + "E.g. deadline breakfast /by 2022-12-31 15:00");
+        }
+
+        // Remove the "deadline " keyword
+        task = task.substring(9);
+
+        // Split the string with /by
+        String[] splitedBy = task.split(" /by ");
```
simkianseng, line 209: Seems like the variable `splitedBy` was misspelled, perhaps can refer to W4.6c of the notes [here](https://nus-cs2103-ay2324s2.github.io/website/book/codeQuality/nameWell/basic/useStandardWords/). This was found in other areas of the codebase as well.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,220 @@
...
+            // File exist but not exist
+            if (i == splitedFile.length - 1 && Files.isDirectory(path)) {
+                throw new DukeException();
+            }
+        }
+
+        Path path = Paths.get(home, this.filePath);
+
+        // Read from file
+        try {
```
simkianseng, line 95: It seems that your code have deep nesting. Perhaps the code block can be simplified? Here is an example of a deeply nested (or arrowhead style) code:
```java
int subsidy() {
    int subsidy;
    if (!age) {
        if (!sub) {
            if (!notFullTime) {
                subsidy = 500;
            } else {
                subsidy = 250;
            }
        } else {
            subsidy = 250;
        }
    } else {
        subsidy = -1;
    }
    return subsidy;
```
Refer [here](https://nus-cs2103-ay2324s2.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-deep-nesting) for related documentation.

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,220 @@
...
+                        // Date and time
+                        try {
+                            if (splitedFrom.length == 2 && splitedTo.length == 2) {
+                                DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm");
+                                LocalDateTime parsedFromDateTime = LocalDateTime.parse(from, formatter);
+                                LocalDateTime parsedToDateTime = LocalDateTime.parse(to, formatter);
+                                Task newTask = new Event(content, isDone, parsedFromDateTime, parsedToDateTime);
+                                items.add(newTask);
+                            } else if (splitedFrom.length == 2 && splitedTo.length == 1) {
+                                DateTimeFormatter formatter1 = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm");
```
simkianseng: Perhaps can the variable name `formatter1` and `formatter2` be something more meaningful? Maybe something like `formatterToDate` and `formatterFromDate` respectively? Refer [here](https://nus-cs2103-ay2324s2.github.io/website/schedule/week4/topics.html#implementation-code-quality-naming-intermediate-use-name-to-explain) for more documentation.

src/main/java/duke/task/Event.java
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    /**
+     * Returns various fields of the task.
+     *
+     * @return A string array containing the description, done status, and formatted due date and time of a task.
+     */
+    @Override
+    public String[] getFields() {
+        String[] result = new String[4];
```
simkianseng: Perhaps the variable `result` can be in plural?
> Plural form should be used on names representing a collection of objects. &ndash; se-education.org

Here is an example:
```java
Collection<Point> points;
int[] values;
```
Refer for the documentation [here](https://se-education.org/guides/conventions/java/intermediate.html).

simkianseng (commented):
Minor code quality issues. Overall good codebase. Good job! 😄

data/tasks.txt
PallonCX: If you are referring to the file that stores the tasks, actually I'm also considering to add them to gitignore. As far as I'm concerned, we are not given explicit instructions on what should be added to gitignore, please tell me if you know where to get more information about this.

src/main/java/duke/Storage.java
PallonCX: My concern is regarding the file path given by user, e.g. for a file path `folderA/folderB/fileC`, indeed System.getProperty("user.dir") already ensures that the path it returns all exists, but when I append the file path given by the user, it could happen certain errors such as:

1. folderA and / or folderB does not exist or exist as a file with same name
2. fileC does not exist or exist as a folder with same name

Therefore, I decide to handle this potential problem by performing checks on each layer, and create a file at the given file path iff there is no error as mentioned above.

src/main/java/duke/Storage.java
PallonCX: That is a good catch, will fix it later, thanks!

src/main/java/duke/Parser.java
PallonCX: I'm not sure whether it is a common practise to put enum as a class of its own or make it part of some other class, but sounds like a valid suggestion as enums are kind of similar to classes, thanks!

src/main/java/duke/command/Command.java
PallonCX: Ah yes

src/main/java/duke/command/DeleteCommand.java
PallonCX: Ooh yea thanks!

src/main/java/duke/task/Deadline.java
PallonCX: I choose to create a new method named `getDescriptionStatus` instead of using the original `toString` inherit from Object class because I'm not sure how many customised strings I will need to make for the class along the way in future increments.

src/test/java/duke/task/DeadlineTest.java
PallonCX: Oh yea you are right, I'm so frustrated while setting up the Gradle and making the JUnit tests, until a point that I only want it to work, which makes me forget the naming convention haha.

src/main/java/duke/command/AddCommand.java
PallonCX: Oww alright, I saw from somewhere (can't recall) that we can use some tag, something like @inherit..., but I didn't manage to find it in my IDE. Any clue about it?

src/main/java/duke/DukeException.java
PallonCX: Ooh yea you are right!

src/main/java/duke/Parser.java
PallonCX: Hmmm my original meaning for the variable is "splited string by the keyword '/by'". What do you think a better name will be?

src/main/java/duke/Storage.java
PallonCX: Hmm it's quite hard for me to simplify it due to the design of my code. To be more precise:
1. The `try` block is required to catch exceptions.
2. The `while` block is required to read inputs continuously.
3. The `for` block is required to consider every task.
4. The `switch` block is required to decide which command to perform.
5. Some `try` block deep inside is indeed redundant as there is a big `try` block at very outside.
6. Some `if-else` block is required to check for specific condition.

Feel free to tell me if you have any specific suggestion, thanks!

src/main/java/duke/Storage.java
PallonCX: Sounds like a valid one, I feel too tedious when setting up these part hahaha. Thanks!

src/main/java/duke/task/Event.java
PallonCX: Ah true!

src/main/java/duke/Parser.java
simkianseng: I think the past tense of "split" is "split". I don't think there is such a word as "splited". Hence the variable name can be called `splitBy`, but no worries just a minor issue.

src/main/java/duke/Storage.java
simkianseng: 1. Perhaps combine multiple `try...catch` blocks into one? I also noticed that you catch `Exception` instead of specifying the specific exception to catch (e.g. `IOException` or `FileNotFoundException`) which is something we what to avoid (I subconsciously remembered in one of the briefing video @damithc mentioned something along the lines of "Don't catch them all" but I can't seem to find it in the textbook). Also, take note to catch all checked exceptions (refer to W1.5 [here](https://nus-cs2103-ay2324s2.github.io/website/schedule/week1/topics.html#W1-5)) The following example might also fix your issues.
```java
try {
    <your code>
} catch (IOException e) {
    <your code>
} catch (FileNotFoundException e) {
    <your code>
}
```
2./3./4./5./6. Perhaps the blocks can be abstracted into smaller methods.

Not really sure if mine is a good example but maybe you can take some inspiration [here](https://github.com/SimKianSeng/ip/blob/master/src/main/java/simpli/storage/Storage.java)?

As everyone has different implementations it's quite hard for me to tell you what is right, but I can point out what doesn't seem right to me and maybe provide you with the reference materials. Hope this was helpful.

data/tasks.txt
tankh99: Apologies, this wasn't part of the course. I was just giving advice in general

src/main/java/duke/Storage.java
tankh99: You could just append the path to user.dir and then check use File.exists() method once to see it's valid. A for loop is most likely unnecessary here

src/main/java/duke/command/AddCommand.java
tankh99: I dont think such an annotation exists? You just don't have to write any comment at all for overridden methods

data/tasks.txt
PallonCX: No bro please don't apologise, I feel so bad for it hahaha. I'm really curious to know what should be gitignore while I'm doing the works. Since you mention that it wasn't part of the course, perhaps I will ask at forum to get some general advices from the teaching team. This point that you have brought up is absolutely great and useful, thanks a lot!

src/main/java/duke/Storage.java
PallonCX: Thanks for your nice elaboration, I will try to apply them in my code!

PR #343, opened 2024-02-01
[vannykin] iP

vannykin:
LumiereMax frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,

2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [x]  Managing tasks

```
public class Main {
    }
}
```


src/main/java/Deadline.java
```diff
@@ -0,0 +1,28 @@
...
+    private String byWhen;
+    private LocalDate deadline;
+
+    public Deadline(String item, boolean marked, String byWhen) {
+        super(item, marked);
+        this.byWhen = byWhen;
+        this.deadline = LocalDate.parse(byWhen);
+    }
+
+    public String getbyWhen() {
```
ZHANGTIANYAO1: Should this be getByWhen?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,28 @@
...
+        return this.byWhen;
+    }
+
+    @Override
+    public String stringify() {
```
ZHANGTIANYAO1: What does `stringify `mean? I know there is a stringify fun in JavaScript for JSON things, does it also defined in java?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,28 @@
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+public class Deadline extends Task {
+    private String byWhen;
+    private LocalDate deadline;
+
+    public Deadline(String item, boolean marked, String byWhen) {
```
ZHANGTIANYAO1: Maybe use `isMarked`?

src/main/java/Lumiere.java
```diff
@@ -0,0 +1,163 @@
+import java.util.*;
```
ZHANGTIANYAO1: Might be better to avoid using * to import everything in the package.

src/main/java/Task.java
```diff
@@ -0,0 +1,29 @@
+public class Task {
+    private boolean marked;
```
ZHANGTIANYAO1: Should this be `isMarked`?

src/main/java/lumiere.txt
```diff
@@ -0,0 +1,2 @@
+T | false | finish up part 4
+D | false | decide on rs | 2024-02-29
```
ZHANGTIANYAO1: May not need to upload this file?

ZHANGTIANYAO1 (commented):
Overall LGTM 👍 
Just some small nits to fix.

src/main/java/Deadline.java
angsongyee: I think the method they were looking for might be `toString()`? Based on how this method is used I think that would be a more appropriate name. Also, another benefit of overriding `toString()` instead of writing an entirely new method is because then Java can automatically call it whenever the object needs to be converted to a String (i.e. string concatenation or passing objects as arguments to `System.out.println()`).
```suggestion
    public String toString() {
```

src/main/java/Deadline.java
```diff
@@ -0,0 +1,28 @@
...
+    @Override
+    public String stringify() {
+        String m = "";
+        if (super.isMarked())
+            m = "[X]";
+        else
+            m = "[ ]";
+        return "[D]" + m + " " + super.stringify() + " (by: "
+                + this.deadline.format(DateTimeFormatter.ofPattern("MMM d yyyy")) + ")";
+    }
```
angsongyee: It's not necessary to use the keyword `super` here, and I think it makes the code harder to read. On the other hand, `super.stringify()` is definitely an appropriate usecase for it so good job for that👍. Basically, you should only use `super` when you need the superclass's implementation of the method instead of the one in your current class.
```suggestion
        if (isMarked())
            m = "[X]";
        else
            m = "[ ]";
        return "[D]" + m + " " + super.stringify() + " (by: "
                + this.deadline.format(DateTimeFormatter.ofPattern("MMM d yyyy")) + ")";
    }
```

src/main/java/Lumiere.java
```diff
@@ -0,0 +1,163 @@
...
+import java.nio.file.Paths;
+
+public class Lumiere {
+    private static String fileLocation = "./lumiere.txt";
+    private static Path filePath = Paths.get(fileLocation);
+
+    public static void main(String[] args) {
+        Scanner sc = new Scanner(System.in);
+
+        List<Task> list = loadTasksFromFile();
```
angsongyee: Maybe a more descriptive name could be used here? 
```suggestion
        List<Task> tasks = loadTasksFromFile();
```

src/main/java/Lumiere.java
```diff
@@ -0,0 +1,163 @@
...
+        System.out.println("Hello! I'm Lumiere\n" + "Tell me what I can do for you, be my guest!\n");
+
+        while (true) {
+            String command = sc.nextLine();
+            if (command.equals("list")) {
+                System.out.println("Here are the tasks in your list:");
+                for (int i = 0; i < list.size(); i++) {
+                    System.out.println(Integer.toString(i + 1) + "." + list.get(i).stringify());
```
angsongyee: As mentioned above, if you override `toString()` instead of using `stringify()` you should be able to print your tasks like this.
```suggestion
                    System.out.println(Integer.toString(i + 1) + "." + list.get(i));
```

src/main/java/Todo.class
angsongyee, line 1: In addition to `lumiere.txt`, `.class` files do not need to be committed to the repo. You should try to modify your `.gitignore` to exclude these types of files from your commits.

angsongyee (commented):
Code looks good, there's only a few minor style issues. Keep up the good work 👍.

PR #350, opened 2024-02-01
[Giselle Tan] iP

GiselleTan:
### Emis

> "The secret of getting ahead is getting started. The secret of getting started is breaking your complex overwhelming tasks into small manageable tasks, and then starting on the first one." - Mark Twain ([source](https://www.goodreads.com/quotes/219455-the-secret-of-getting-ahead-is-getting-started-the-secret))

Emis keeps track of everything going on in your life. It's
- text-based
- easy to learn
- fast to use

All you need to do is:
1. download it from [here](https://github.com/GiselleTan/ip/releases/tag/v0.1)
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you! 🤩

And it is **FREE!**

Features:
- [ ] Managing tasks
- [ ] Reminders (coming soon)

```java
public class Main {
    }
}
```


src/main/java/emis.command/Command.java
```diff
@@ -0,0 +1,7 @@
+package main.java.emis.command;
+import main.java.emis.*;
```
gavin331: Perhaps it will be better to specify the files that are imported instead of using the * to import everything.

src/main/java/emis.task/Deadline.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    @Override
+    public String storeString() {
+        return "D | " + super.storeString() + " | " + this.by;
+    }
+
+    @Override
+    public String toString() {
+        return"[D]" + super.toString() + " (by: " + this.doByDateTime.format(DateTimeFormatter.ofPattern("MMM dd yyyy HH:mm")) + ")";
```
gavin331: Perhaps line wrapping should be used here since it is a long line of code.

src/main/java/emis/Storage.java
```diff
@@ -0,0 +1,91 @@
...
+        this.al = new ArrayList<>();
+    }
+
+    public ArrayList<Task> load() throws EmisException {
+        try {
+            Scanner sT = new Scanner(new File(this.filePath));
+            while (sT.hasNextLine()) {
+                // read from file, add to arraylist al
+                String input = sT.nextLine();
+                String[] data = input.split("\\|");
```
gavin331: Should the name for String[] be in singular form or plural? Same for the ArrayList<Task> al above

src/main/java/emis/Storage.java
```diff
@@ -0,0 +1,91 @@
...
+                String taskType = data[0].trim();
+                String retrieveStatus = data[1].trim();
+                boolean convertStatus = retrieveStatus.equals("1");
```
gavin331: Perhaps "convertStatus" can be renamed to sound more like a boolean

gavin331 (commented):
Overall the code looks good except for a few coding standard violations. For eg. naming violations.

src/main/java/emis.command/Command.java
tanguoyang: would be good to import as needed!

src/main/java/emis.command/DeadlineCommand.java
```diff
@@ -0,0 +1,24 @@
+package main.java.emis.command;
+import main.java.emis.*;
+import main.java.emis.task.Deadline;
```
tanguoyang: Are one of these imports redundant?

src/main/java/emis.command/ToDoCommand.java
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    @Override
+    public void execute(TaskList t, Ui ui, Storage s) {
+        t.addTask(new ToDo(this.description));
+        s.updateStorage();
+    }
+
+    @Override
+    public boolean isExit() {
```
tanguoyang: I wonder if theres a better way of implementing this? As the number of commands grow, every command will have an isExit(), but yet only one command will ever be isExit() = true?

src/main/java/emis.task/Task.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public String getStatusIcon() {
+        return (isDone ? "X" : " ");
+    }
+
+    public void markAsDone() {
+        this.isDone = true;
+        Ui.showLine();
+        System.out.println("\tNice! I've marked this task as done:");
+        System.out.println("\t\t[" + this.getStatusIcon() + "] " + this.description);
```
tanguoyang: I wonder if these lines can be in Ui class as well since they are showing things to the user?

src/main/java/emis/Parser.java
```diff
@@ -0,0 +1,64 @@
...
+            if (fullCommand.equals("bye")) {
+                return new ExitCommand();
+            } else if (fullCommand.equals("list")) {
+                return new PrintCommand();
+            } else {
+                throw new EmisException("Invalid command.");
+            }
+        } else {
+            // spaces found
+            // possible commands are "todo", "deadline", "event", "mark", "unmark", "delete"
```
tanguoyang: an enum here would be really useful!

`enum Actions {
	TODO,
	DEADLINE,
	EVENT
}`

src/test/java/TaskListTest.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.ArrayList;
+
+import main.java.emis.exceptions.EmisException;
+import main.java.emis.task.Task;
+import main.java.emis.TaskList;
+import main.java.emis.task.ToDo;
+
+public class TaskListTest {
+    @Test
+    public void markAsDone_exceptionThrown() throws EmisException {
```
tanguoyang: perhaps the naming could be more specific, marking what as done leads to to exception being thrown?

maybe a better name would be
markIndexZeroAsDone_exceptionThrown

src/test/java/TaskListTest.java
```diff
@@ -0,0 +1,45 @@
...
+        try {
+            testTasklist.markAsDone(0);
+            fail(); // the test should not reach this line
+        } catch (EmisException e) {
+            assertEquals("This task does not exist!", e.getMessage());
+        }
+    }
+
+    @Test
+    public void deleteTask_exceptionThrown() {
```
tanguoyang: same concern here! there may be times where deleteTask shouldnt always cause exception thrown so lets try to be more specific with the naming!

tanguoyang (commented):
some minor concerns, but keep up the good work!

PR #374, opened 2024-02-01
[dhlee03] iP

dhlee03:
### FriendlyTool

> "For every minute spent organising, an hour is earned." – Benjamin Franklin([source](https://www.theplainsimplelife.com/quotes-about-being-organised/))

FriednlyTool helps you to organise tasks easily and save time.

**Why** using it? 

- Easy and Fast to use
- Text-based 
- User-friendly features😉

**Features**:

- [x]  Adding/Managing tasks
- [x]  Viewing a list of tasks
- [x]  Finding tasks

**How** to install?

1. Download it from [here](https://github.com/dhlee03/ip/releases)
2. Open it
3. Add your own task
4. Enjoy its features

And it's _Free_!

```java

public class Main {
    }
}
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,283 @@
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.*;
+
 public class Duke {
+    private boolean isActive;
+    private List<Task> myList;
```
yashpola: This might be better if it was `tasks` as its more descriptive of what the list is

src/main/java/Duke.java
```diff
@@ -1,10 +1,283 @@
...
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        Duke ft = new Duke();
+        ft.init();
+    }
+    public void init() {
```
yashpola: A more descriptive name might be `greetUser()` as its specific to the context of the class and the client knows what action is triggered by calling this method rather than if it was simply `init`

src/main/java/Duke.java
```diff
@@ -1,10 +1,283 @@
...
+            } catch (ftException e) {
+                System.out.println(e.getMessage());
+            }
+        }
+        System.out.println("    ____________________________________________________________\n"
+                + "    Bye. Hope to see you again soon!\n"
+                + "    ____________________________________________________________" );
+    }
+
+    private void nextAction(String input) throws ftException {
```
yashpola: While `nextAction` does imply a continuation of interaction with the user, something like `handleUserResponse` might be better as it makes it clear who is taking the next action (the bot in this case) and who it is meant for (the user)

src/main/java/Duke.java
```diff
@@ -1,10 +1,283 @@
...
+                    break;
+                default:
+                    throw new ftException("Unknown Command: Please use a correct command");
+            }
+        }catch (IllegalArgumentException e) {
+            throw new ftException("Unknown Command: Please use a correct command");
+        }
+    }
+
+    private void addTask(StringTokenizer st, CommandTypes ct) throws ftException {
```
yashpola: This codefile is a bit long for me, I think it would benefit from more modularization through perhaps a Parser class or something to handle the parsing of the request and task creation. This also reduces the repeated `switch-case` to check task type before acting on the user request

src/main/java/Duke.java
```diff
@@ -1,10 +1,283 @@
...
+            System.out.println("    ____________________________________________________________\n"
+                    + "    Nice! I've marked this task as done:\n"
+                    + "      " + task.toString()
+                    + "\n    ____________________________________________________________\n");
+        } else {
+            throw new ftException("Error: Please provide valid index");
+        }
+    }
+
+    private void unmark(StringTokenizer st) throws ftException {
```
yashpola: Since your other methods are in the form of `deleteTask` and `updateTask` etc, to be consistent it would be better if these were `unmarkTask` and `markTask` as well

yashpola (commented):
The functionality and naming is mostly clear despite the length of the files but for future readability and expansion, my biggest suggestion would be more modularisation most evidently through a class dedicated to string parsing

PR #375, opened 2024-02-01
[Ty-stan0417] iP

Ty-stan0417:


src/main/java/MarkNothingException.java
```diff
@@ -0,0 +1,5 @@
+public class MarkNothingException extends Exception {
```
tanjieling0: Perhaps there could be a better name for this exception? perhaps InvalidTaskException or something 😬

src/main/java/Tyler.java
```diff
@@ -0,0 +1,189 @@
...
+                      System.out.println("      " + newTask.toString());
+                      curr++;
+                      System.out.println("    Now you have " + curr + " tasks in the list");
+
+                  } else if (arr[0].equals("deadline")) {
+                      if (arr.length < 2) {
+                          throw new EmptyNameException();
+                      }
+                      String[] rest1 = input.split(" ", 2);
+                      String[] rest2 = rest1[1].split(" /by ");
```
tanjieling0: Could there be a better way to differentiate the variables instead of 1 and 2? The same naming convention can be seen in other parts of the code as well.

src/main/java/Tyler.java
```diff
@@ -0,0 +1,189 @@
...
+        } catch (FileNotFoundException e) {
+            File file = new File(FILE_PATH);
+            file.createNewFile();
+        }
+    }
+
+    public static String toFileString(Task task) {
+        String name = task.name;
+        boolean isDone = task.isDone;
+        int isDoned = isDone ? 1 : 0;
```
tanjieling0: Maybe the "is" variable naming should be reserved for boolean values whereas in this case it is an integer

src/main/java/EmptyNameException.java
```diff
@@ -0,0 +1,5 @@
+public class EmptyNameException extends Exception {
```
tanjieling0: Nice naming :)

tanjieling0 (commented):
LGTM, code is fairly neat and tidy, but perhaps renaming some of the variables / classes can help make the code more readable for other code reviewers :)

Ty-stan0417:
About this A-Packages, I don't know why the Tyler directory and Command directory start with capital letter. Actually in my code, it is small letter, all my package is in small letter.

FangRuoqing (commented):
Your choice of names throughout the codebase is consistent and easy to understand, which is awesome for anyone reading your code. Maybe adding some comments for your methods and classes would be more helpful? Generally, I like how your code is nicely organized into packages, making it a breeze to find what you need.

src/main/java/tyler/command/EventCommand.java
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) {
+        Event newTask = new Event(this.taskName, this.start, this.end);
+        tasks.addTask(newTask);
+        int num = tasks.getNumOfTasks();
+        ui.showTaskAdded(newTask, num);
+    }
+}
```
redcolorbicycle, line 48: variable names are aptly named and logic is easy to follow.

src/main/java/tyler/parser/Parser.java
```diff
@@ -0,0 +1,64 @@
+package tyler.parser;
+import tyler.command.*;
+import tyler.exception.*;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class Parser {
```
redcolorbicycle, line 28: Perhaps you could consider splitting up the function - would grouping the code into a function from to be handled by another class make it a little less lengthy?

src/main/java/tyler/parser/Parser.java
```diff
@@ -0,0 +1,64 @@
...
+            return new TodoCommand(name);
+        } else if (command.equals("deadline")) {
+            if (input.length < 2) {
+                throw new EmptyNameException();
+            }
+            String[] info = line.split(" ", 2)[1].split("/by ");
+            String name = info[0];
+            LocalDateTime end = LocalDateTime.parse(info[1], INPUT_DATE_FORMAT);
+            return new DeadlineCommand(name, end);
+        } else if (command.equals("event")) {
```
redcolorbicycle: Instead, would 
```
} else if (command.equals("event")) {
HandleCommand.HandleEvent;
}
```
abstract away these details and make it easier to reference? 

I like that the constants are in SCREAMING_SNAKE_CASE, classes in PascalCase and variables in camelCase!

src/main/java/tyler/storage/Storage.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.Scanner;
+
+public class Storage {
+    protected String filePath;
+    private final DateTimeFormatter OUTPUT_DATE_FORMAT = DateTimeFormatter.ofPattern("dd MMM yyyy h:mm a");
+    public Storage(String filePath) {
+        this.filePath = filePath;
+    }
+
+    public ArrayList<Task> load() throws IOException {
```
redcolorbicycle, line 36: Similar to above, the code logic is sound and the naming convention is well done and easy to follow! Could abstracting some details into function make it even better?

src/main/java/tyler/ui/Ui.java
```diff
@@ -0,0 +1,57 @@
+package tyler.ui;
+import tyler.task.Task;
+import tyler.task.TaskList;
+
+import java.util.Scanner;
+import java.util.ArrayList;
+
+public class Ui {
```
redcolorbicycle, line 12: Code follows the coding standard well!

redcolorbicycle (commented):
Overall, code is named according to coding standard, is easy to understand and well written!  Some cosmetic suggestions to make reading it a little easier have been requested.
I think the code is well written! :+1:

src/main/java/MarkNothingException.java
Ty-stan0417: It's a good suggestion. I already name it InvalidTaskException. Thanks.

src/main/java/Tyler.java
Ty-stan0417: Good suggestion. I have change the naming so that can understand what the code is through the name.

src/main/java/Tyler.java
Ty-stan0417: Good suggestion. I will change it later.

PR #381, opened 2024-02-02
[chin-herng] iP

chin-herng:
# A first-level heading
> Text that is a quote. This site was built using [GitHub Pages](https://pages.github.com/).
- George Washington
* John Adams
+ ~~This was mistaken text~~ _This text is italicized_
1. James Madison
2. James Monroe
3. John Quincy Adams :+1:
- [x] #739
- [ ] https://github.com/octo-org/octo-repo/issues/740
- [ ] Add delight to the experience when all tasks are complete :tada:

Use `git status` to list all new or modified files that haven't yet been committed.

```ruby
require 'redcarpet'
markdown = Redcarpet.new("Hello World!")
puts markdown.to_html
```


src/main/java/bob/Bob.java
```diff
@@ -0,0 +1,38 @@
+// TODO: divide into multiple packages
+package bob;
+
+import java.time.format.DateTimeParseException;
+
+public class Bob {
+    public static void handleMark(int taskIndex, boolean isDone) throws InvalidTaskIndexException {
```
miffi: I feel that the `handle` prefix is redundant. `mark` is a verb already, and it describes what is being done. Therefore, maybe returning mark will make it more concise. If the usage of the function is to be made clearer,  I think maybe something like `markTask` would be better. (Then you could also change taskIndex to just index as it would be pretty obvious what it's indexing imo)

Similar sentiment for the other functions this class.

src/main/java/bob/Parser.java
```diff
@@ -0,0 +1,134 @@
...
+    public static final String EXIT = "exit";
+    public static final String LIST = "list";
+    public static final String MARK = "mark";
+    public static final String UNMARK = "unmark";
+    public static final String TODO = "todo";
+    public static final String DEADLINE = "deadline";
+    public static final String EVENT = "event";
+    public static final String DELETE = "delete";
+
+    public static final DateTimeFormatter INPUT_DATE_FORMATTER = DateTimeFormatter.ISO_LOCAL_DATE;
```
miffi: I've seen this constant being used for both input and output, so maybe just `DATE_FORMATTER` might be better.

src/main/java/bob/Event.java
```diff
@@ -0,0 +1,26 @@
...
+    protected LocalDate to;
+
+    public Event(String description, String from, String to) throws DateTimeParseException {
+        super(description);
+        this.from = LocalDate.parse(from, Parser.INPUT_DATE_FORMATTER);
+        this.to = LocalDate.parse(to, Parser.INPUT_DATE_FORMATTER);
+    }
+
+    public String format() {
+        return "event | " + super.format() + " | " + this.from.format(Parser.INPUT_DATE_FORMATTER)
```
miffi: Okay this line has thrown me in a bit of a loop. You have used both INPUT and OUTPUT for output in this case so I'm not really sure what different significance they have. Maybe you can try changing the names or splitting into other variables to make this more obvious?

src/main/java/bob/Ui.java
```diff
@@ -0,0 +1,89 @@
...
+    public static final String LIST_HEADER = "list of tasks:";
+
+    public static final String MARK_HEADER = "good job!";
+    public static final String UNMARK_HEADER = "ok you just undid this task";
+
+    public static final String DELETE_HEADER = "removed:";
+    public static final String DELETE_FOOTER = "%d task(s) left";
+
+    public static final String EMPTY_DESCRIPTION = "missing description for %s";
+    public static final String INVALID_COMMAND = "what";
```
miffi: Honestly having all of the `INVALID`s in the UI is an interesting choice. It brings all of the CLI text into one place which would make it easier to edit and understand how the text looks as a whole. However, it does make understanding what the exceptions say when reading the code. I don't have a particular recommendation, but imo having the messages where they're used, i.e. the exceptions, would be better.

src/main/java/bob/Deadline.java
```diff
@@ -0,0 +1,23 @@
...
+public class Deadline extends Task {
+
+    protected LocalDate by;
+
+    public Deadline(String description, String by) throws DateTimeParseException {
+        super(description);
+        this.by = LocalDate.parse(by, Parser.INPUT_DATE_FORMATTER);
+    }
+
+    public String format() {
```
miffi: I see that this is overriding the superclass. Maybe you could add @Override to this and to the same function in other subclasses to make this obvious.

src/main/java/bob/Parser.java
```diff
@@ -0,0 +1,134 @@
...
+
+    public static String[] extractParameters(String parametersString,
```
miffi: It's a bit hard to understand what the parameters do. From the code it seems that the `parametersString` is the input to extract the parameters from, but I don't see any use for parameters other than its length. Adding a javaDoc header might help. If the only use for parameters is the length I think that passing the length itself is better.

src/main/java/bob/Parser.java
```diff
@@ -0,0 +1,134 @@
...
+            }
+            result[i + 1] = splitString[1];
+        }
+
+        result[0] = splitString[0];
+
+        return result;
+    }
+
+    public static void processDeleteOrMarkCommands(String[] commandArgs) throws NumberFormatException,
```
miffi: `Process` is vague. It doesn't really indicate what the `processing` is. I think changing to `make` or maybe `create` would describe what this is doing better.

src/main/java/bob/Parser.java
```diff
@@ -0,0 +1,134 @@
...
+                    Bob.handleAdd(commandArgs[0], parameters);
+            }
+        } catch (ParameterNotFoundException e) {
+            Ui.print(e.getMessage());
+        } catch (DateTimeParseException e) {
+            Ui.print(Ui.INVALID_DATE_FORMAT);
+        }
+    }
+
+    public static boolean listen() {
```
miffi: I think this name is fine. It describes what the command is doing. However, it is not obvious what it is supposed to do from just its name. A javaDoc header imo should help with that.

src/main/java/bob/Parser.java
```diff
@@ -0,0 +1,134 @@
...
+        case Parser.DELETE:
+            // Fallthrough
+        case Parser.UNMARK:
+            // Fallthrough
+        case Parser.MARK:
+            try {
+                processDeleteOrMarkCommands(commandArgs);
+            } catch (NumberFormatException e) {
+                // The more "correct" way is to throw an InvalidTaskIndexException?
+                Ui.print(String.format(Ui.INVALID_TASK_INDEX, commandArgs[1]));
```
miffi: Instead of catching the `NumberFormatException` and printing out the index here, I feel that it might be better to have another exception whose message itself has the index with the specific command. That would would express the intent of the exception in the code better imo.

miffi (commented):
Hello :). I'm a peer reviewer for your iP! I'm doing my second review for you (the code quality one). I have some recommendations, but overall I think that your code is readable and well structured.

One overarching suggestion is that you could move Task and its subclasses into a bob.task package. I feel like that would make their relation more obvious.

src/main/java/bob/Bob.java
chin-herng: hi! im sort of establishing a convention that whatever commands has to go through the Bob class first, and Bob will coordinate everything - he "handles" the command over to other classes. for example Bob "handles" the mark command by telling TaskList to actually do the marking

hope it makes sense haha, that's my justification for why every method in Bob has a handle prefix, it's to emphasise that the method is a command handler.

src/main/java/bob/Parser.java
chin-herng: hi! the input and output date format are different. the tasks read from the command line, and are stored in the hard disk in the format of input date format. the tasks are being printed in the format of output date format.

src/main/java/bob/Event.java
chin-herng: yea youre right that the naming can be better. the format method is to format the task in the format stored in the hard disk. input format is used so that when loading the tasks from the hard disk, we can load the tasks as if we are reading the tasks in the file.

src/main/java/bob/Ui.java
chin-herng: thanks! i considered putting these messages in the corresponding exceptions but i was thinking that it's more convenient to customise my bot if i put all the ui-related outputs in the Ui class. there might be alternative ways to do this and i might do the refactor in the future.

src/main/java/bob/Parser.java
chin-herng: yea, this method deserves a javadoc, will add it in the future!

src/main/java/bob/Parser.java
chin-herng: youre right that the word process is chosen due to lack of vocabulary. on second thought it might be better to use "parse".

src/main/java/bob/Parser.java
chin-herng: thanks for the suggestion!

src/main/java/bob/Bob.java
```diff
@@ -0,0 +1,47 @@
...
+        boolean isExit = false;
+        while (!isExit) {
+            try {
+                String fullCommand = ui.readCommand();
+                ui.showLine();
+                Command command = Parser.parse(fullCommand);
+                command.execute(ui, storage, tasks);
+                isExit = command.isExit();
+            } catch (BobException e) {
+                ui.show(e.getMessage());
```
Greg-Tan: great usage of try and catch block to handle your errors

src/main/java/bob/Parser.java
```diff
@@ -0,0 +1,182 @@
...
+import bob.command.AddDeadlineCommand;
+import bob.command.AddEventCommand;
+import bob.command.AddTodoCommand;
+import bob.command.Command;
+import bob.command.DeleteCommand;
+import bob.command.ExitCommand;
+import bob.command.ListCommand;
+import bob.command.ListDueInCommand;
+import bob.command.ListOnDateCommand;
+import bob.command.MarkCommand;
```
Greg-Tan, line 19: should arrange your imports in alphabetical order if possible

src/main/java/bob/Parser.java
```diff
@@ -0,0 +1,182 @@
...
+            = DateTimeFormatter.ofPattern(DATE_PATTERN);
+
+    private static final String DATETIME_PATTERN = DATE_PATTERN + " HHmm";
+    private static final DateTimeFormatter DATETIME_FORMATTER
+            = DateTimeFormatter.ofPattern(DATETIME_PATTERN);
+
+    private static String[] extractParameters(String parametersString,
+                                             String[] parameters) throws ParameterNotFoundException {
+        // TODO: generalise the method to any given parametersString, desiredParameters and separator
+        // TODO: perform ParameterNotFoundException checks correctly
```
Greg-Tan: should remove comments before making the pull request

src/main/java/bob/Bob.java
```diff
@@ -0,0 +1,47 @@
...
+            }
+        }
+    }
+
+    private void run() {
+        ui.showWelcome();
+        runCommandLoopUntilExitCommand();
+    }
+
+    public static void main(String[] args) {
```
Greg-Tan: exceptional OOP design to your chatbot!

Greg-Tan (commented):
LGTM. Please remember to remove unnecessary comments before making your pull requests :)

src/main/java/bob/Bob.java
chin-herng: sflr thanks!

src/main/java/bob/Parser.java
chin-herng: sflr, but isn't the highlighted imports already in alphabetical order?

src/main/java/bob/Parser.java
chin-herng: sflr, I do think that these comments are helpful to communicate the loose ends to tie before the PR can be merged, or possible improvements after the PR is merged, which is why I left it in the PR.

src/main/java/bob/Bob.java
chin-herng: sflr thanks!

PR #384, opened 2024-02-02
[AdityaB4] iP

AdityaB4:
# Chimp 🦧
> "HOO-HOO" - chimp

### Features
1. Adding tasks
2. Finding tasks
3. Deleting tasks

### Steps to use
- Create a `.jar` file
- Use the **jar** file

### Features
- [x] Managing tasks

More information [here](https://github.com/nus-cs2103-AY2324S2/ip/pull/384)

```java
System.out.println("This is fenced code and I have nothing to put here");
```


src/main/java/chimp/command/DeadlineCommand.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.time.LocalDate;
+
+import chimp.core.Storage;
+import chimp.core.TaskList;
+import chimp.core.Ui;
+import chimp.exception.CommandExecuteException;
+
+public class DeadlineCommand extends Command {
+    String text;
```
bennyLCK: Perhaps a more meaningful name like taskName would make more semnantic meaning in command classes in general?

src/main/java/chimp/command/DeleteCommand.java
```diff
@@ -0,0 +1,31 @@
...
+    int index;
+
+    public DeleteCommand(int index) {
+        this.index = index;
+    }
+
+    @Override
+    public boolean isExit() {
+        return false;
+    }
```
bennyLCK, line 32: I like how you applied the concept of overriding suitable parent methods instead of what some people would do, accessing the parent's field isExit, and then manually setting and getting that field.

src/main/java/chimp/command/DeleteCommand.java
```diff
@@ -0,0 +1,31 @@
...
+
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws CommandExecuteException{
+        if (index - 1 < 0 || index - 1 >= tasks.size()) {
```
bennyLCK, line 47: Kuddos to handling the edge cases for invalid integer indexes, but i feel that there can be more error handling for cases where the 'argument' supplied to delete command is say a letter or even empty.

src/main/java/chimp/command/EventCommand.java
```diff
@@ -0,0 +1,32 @@
...
+
+public class EventCommand extends Command {
+    String text;
+    LocalDate fromDate;
+    LocalDate toDate;
+
+    public EventCommand(String text, LocalDate fromDate, LocalDate toDate) {
+        this.text = text;
+        this.fromDate = fromDate;
+        this.toDate = toDate;
```
bennyLCK, line 29: I like your practice of using fully defined naming references for your instance variables!

src/main/java/chimp/command/FindCommand.java
```diff
@@ -0,0 +1,27 @@
+package chimp.command;
+
+import chimp.core.Storage;
+import chimp.core.TaskList;
+import chimp.core.Ui;
+import chimp.exception.CommandExecuteException;
+import chimp.task.Task;
+
+public class FindCommand extends Command{
+    String arg;
```
bennyLCK: Nit: Again, a more meaningful name would allow others to understand your codes better (keyWord for example).

src/main/java/chimp/core/TaskList.java
```diff
@@ -0,0 +1,79 @@
...
+import java.time.LocalDate;
+import java.util.ArrayList;
+
+import chimp.task.Deadline;
+import chimp.task.Event;
+import chimp.task.Task;
+import chimp.task.TaskStatus;
+import chimp.task.Todo;
+
+public class TaskList extends ArrayList<Task>{
```
bennyLCK: Would it be more appropriate to instead encapsulate an `ArrayList<Task>` as a field in `TaskList` class as a `ArrayList<Task>` object is used to represent the internals of the `TaskList` class?

src/main/java/chimp/command/MarkCommand.java
```diff
@@ -0,0 +1,29 @@
...
+    @Override
+    public boolean isExit() {
+        return false;
+    }
+
+    @Override
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws CommandExecuteException {
+        if (num < 1 || num > tasks.size()) {
+            throw new CommandExecuteException("mark must have number argument");
+        }
```
bennyLCK, line 46: Perhaps more edge cases can be considered? This also applies to commands of delete, mark/unmark, find etc. that have a number or keyword as additional info used in the commands. Maybe you can consider cases like  letters/word given instead of an index or when it is empty when you are expecting more info.

src/main/java/chimp/core/Parser.java
```diff
@@ -0,0 +1,96 @@
...
+            case "delete":
+                num = Integer.parseInt(arg.strip());
+                return new DeleteCommand(num);
+            case "bye":
+                return new ExitCommand();
+            case "find":
+                return new FindCommand(arg);
+            default:
+                throw new InvalidCommandException("command \"" + command + "\" is invalid");
+        }
```
bennyLCK: Maybe you should try to implement checkstyle or revisit the coding standards webpage? It is not familiar to us, but the `case` statements should not be indented with respect to the `switch` statement based on our shared coding standard. The IDES normally have auto-styling on which indents for us on save / exit so you will have to be careful and prevent your styling changes from being reverted even after you have removed the indents. On Windows, the shortcut would be to highlight the entire block of `case` statements then press shift + tab to shift the block towards the left.

src/main/java/chimp/core/Storage.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class Storage {
+    /**
+     * Checks if the given command is a save command.
+     *
+     * @param command the command to be checked
+     * @return true if the command is a save command, false otherwise
+     */
+    public static boolean isSaveCommand(String command) {
+        HashSet<String> saveCommands = new HashSet<>(
```
bennyLCK, line 33: I like your use of a `HashSet<String>` to keep track of list of allowed commands in your program, thereby reducing the time needed to retrieve the types of a command from instead maybe a list. But perhaps it would be more well suited to be stored and referenced as a static field in the `Command` class instead so that you would not have to instantiate a `saveCommands` object every time you do a predicate check with this method?

src/main/java/chimp/core/Storage.java
```diff
@@ -0,0 +1,58 @@
...
+            Path directoryPath = filePath.getParent();
+            if (directoryPath != null) {
+                Files.createDirectories(directoryPath);
+            }
+            Files.createFile(filePath);
+        } catch (FileAlreadyExistsException e) {
+            // Ignore this error
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
```
bennyLCK, line 63: I like how you separate the actions of creating and writing to a file, but perhaps it would be best to encapsulate the creation of a file as a separate method instead? Especially since we would only be creating a file once at the start.

src/main/java/chimp/core/TaskList.java
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * Adds a new event task to the task list.
+     *
+     * @param task the description of the task
+     * @param from the starting date of the event
+     * @param to the ending date of the event
+     */
+    public void add(String task, LocalDate from, LocalDate to) {
+        super.add(new Event(task, TaskStatus.UNMARKED, from, to));
+    }
```
bennyLCK: I like your decision of overloading the `add` method and rely on it to return the correct type of task during runtime.

src/main/java/chimp/core/TaskList.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Returns a string representation of the TaskList.
+     *
+     * @return a string representation of the TaskList
+     */
```
bennyLCK, line 59: Perhaps the comment here can be omitted entirely since it is trivial? Or the `@return` statement can be omitted here since it is repetitive.

src/main/java/chimp/core/TaskList.java
```diff
@@ -0,0 +1,79 @@
...
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder();
```
bennyLCK, line 62: I like your use of `StringBuilder` here to increase the efficiency of String concatenation.

src/main/java/chimp/core/TaskList.java
```diff
@@ -0,0 +1,79 @@
...
+
+    public Task[] find(String keyword) {
+        ArrayList<Task> matches = new ArrayList<>();
+        for (int i = 0; i < this.size(); i++) {
+            if (this.get(i).toString().contains(keyword)) {
+                matches.add(this.get(i));
+            }
+        }
+        return matches.toArray(new Task[0]);
+    }
```
bennyLCK, line 83: Perhaps you can consider to modify the return type of the method to `ArrayList<Task>` and return `matches` directly? in calling `matches.toArray(new Task[0])`, we would be performing double the work by populating two separate data structures to hold the same data which is wasteful.

src/main/java/chimp/core/Ui.java
```diff
@@ -0,0 +1,101 @@
...
+        phrases.put("greet", greet);
+        phrases.put("bye", bye);
+        phrases.put("mark", mark);
+        phrases.put("unmark", unmark);
+        phrases.put("hoo", hoo);
+        phrases.put("delete", delete);
+        phrases.put("find", find);
+
+        return phrases;
+    }
```
bennyLCK: Perhaps it would be best to store a `HashMap<String, String> phrases` instance as a static field since you would be referring to it whenever you call the `say` method?

src/main/java/chimp/core/Ui.java
```diff
@@ -0,0 +1,101 @@
...
+     * Prints the phrase associated with the given key from the getPhrases hashmap.
+     *
+     * @param key the key to retrieve the phrase from the hashmap
+     */
+    public void say(String key) {
+        String phrase = getPhrases().get(key);
+        System.out.println(DIVIDER);
+        System.out.println(phrase);
+        System.out.println(DIVIDER);
+    }
```
bennyLCK: Perhaps you can consider learning the default `String.format()` method for the Java String class in printing statements with many newline characters, since it can all be grouped up into one print statement?

bennyLCK (commented):
The overall class design LGTM. Overall, I am very delighted that you used various advanced OOP concepts like overloading, overriding and inheritance to a very large extent. However, there were certain coding standard violations too. Firstly, your classes were missing class header comments which is a requirement, secondly, there were certain methods that kept instantiating new instances of data structures which could have otherwise been avoided, lastly, there were some methods which fulfill more than one purpose, perhaps you can consider extracting different functionalities into its individual methods.

src/main/java/chimp/command/DeleteCommand.java
```diff
@@ -19,12 +19,13 @@ public boolean isExit() {
     }
 
     @Override
-    public void execute(TaskList list, Ui ui, Storage storage) throws CommandExecuteException{
-        if (index - 1 < 0 || index - 1 >= list.size())
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws CommandExecuteException{
```
Joshy837: Can add a space after `CommandExecuteException` to follow the code quality standard

src/main/java/chimp/command/UnmarkCommand.java
```diff
@@ -18,11 +18,12 @@ public boolean isExit() {
     }
 
     @Override
-    public void execute(TaskList list, Ui ui, Storage storage) throws CommandExecuteException {
-        if (num < 1 || num > list.size())
+    public void execute(TaskList tasks, Ui ui, Storage storage) throws CommandExecuteException {
```
Joshy837: I like how you used the variable name `num` to indicate that the variable used is an integer

src/main/java/chimp/core/Ui.java
```diff
@@ -4,7 +4,7 @@
 import chimp.task.Task;
 
 public class Ui {
-    private static final String divider = "____________________________________________________________\n";
+    private static final String DIVIDER = "____________________________________________________________\n";
```
Joshy837: I like how you used all caps for `DIVIDER` to explicitly show that it is a constant

Joshy837 (commented):
Overall good quality code and naming conventions, can make slight improvements to make code even clearer!

PR #392, opened 2024-02-02
[Low Zheng Hui] iP

zhenghuil:
# Reacher, helps u remember your tasks to do.
## functions
- add `add`
- edit
1. delete
2. list
for example,
```
add
edit
```
**- [x] mark done**
_- [ ] edit_
- [ ] add
🦅 
> this is a quote
[this is a link](https://github.com/nus-cs2103-AY2324S2/ip/pull/392)


src/main/java/Task.java
```diff
@@ -0,0 +1,25 @@
...
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // mark done task with X
+    }
+
+    public void markDone() {
+        this.isDone = true;
+    }
+    public void markNotDone() {
+        this.isDone = false;
+    }
```
zhuuyicheng: I like that your names representing methods are verbs and written in camelCase, good job following the coding standards.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,14 @@
+import java.util.SplittableRandom;
+
+public class Deadline extends Task {
+    protected String by;
```
zhuuyicheng: I like that your class variables are never declared public, in accordance to the coding standards 👍

src/main/java/Task.java
```diff
@@ -0,0 +1,25 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
```
zhuuyicheng: Good job naming your boolean variable to sound like a boolean as per the coding standards.

src/main/java/Reacher.java
```diff
@@ -0,0 +1,91 @@
...
+                "I'm Reacher.\n" +
+                "Give me tasks.\n" +
+                "Functions are edit, list, delete and bye");
+        while (true) {
+            try {
+                String input = scanner.nextLine();
+                if (input.isEmpty()) {
+                    throw new ReacherException("pls type a task name.");
+                }
+                if (input.equalsIgnoreCase("bye")) { //check for end request
```
zhuuyicheng: While the if/else statements work properly to identify and separate logic blocks for different inputs, maybe you could also look into using switch/case statements or even enums which can look neater since there are so many different commands?

zhuuyicheng (commented):
Overall, LGTM.

src/main/java/Reacher.java
```diff
@@ -0,0 +1,91 @@
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.Scanner;
+
+public class Reacher {
+    public static void main(String[] args) {
+            Scanner scanner = new Scanner(System.in);
+        ArrayList<Task> memory = new ArrayList<>();
```
breezetall: Should these be indented the same?

src/main/java/Reacher.java
```diff
@@ -0,0 +1,91 @@
...
+                if (input.isEmpty()) {
+                    throw new ReacherException("pls type a task name.");
+                }
+                if (input.equalsIgnoreCase("bye")) { //check for end request
+                    System.out.println("bye");
+                    break;
+                } else if (input.equalsIgnoreCase("list")) {// check for list request
+                    System.out.println("Tasks:");
+                    int c = 1;
+                    for (Task task : memory) {
```
breezetall: Perhaps a normal for loop would be better here? Then the declaration of c would be within the same scope as the for loop.

src/main/java/Reacher.java
breezetall: Perhaps a more intuitive name such as taskList?

src/main/java/Reacher.java
```diff
@@ -0,0 +1,91 @@
...
+                            memory.add(t);
+                            System.out.println("I've added " + t.toString());
+                        } else {
+                            throw new ReacherException("That is not a type of task.");
+                        }
+                    } catch (ReacherException e) {
+                        System.out.println(e.getMessage());
+                    }
+                }
+            } catch (Exception e) {
```
breezetall: I like that you added a catch Exception to catch all the exceptions that might have gone past ReacherException!

breezetall (commented):
Overall, I think this code is well-written. Just some minor coding standard and naming issues.

PR #394, opened 2024-02-02
[Hibeom0929] iP

Hibeom0929:
# UnivusPro
> "Your mind is for having ideas, not holding them." - David Allen ([source](https://dansilvestre.com/productivity-quotes/))

UnivusPro frees your mind of having to remember things you need to do. It's,
- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,
1. download if from [here](https://github.com/Hibeom0929/ip/releases/tag/A-Jar).
2. double-click it.
3. add your tasks
4. let it manage your tasks for you :D

And it is **FREE**!

Features:
- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon)

#
```
public class Main {
    public static void (String[] args) {
    }
}
```


src/main/java/Deadlines.java
```diff
@@ -0,0 +1,13 @@
+public class Deadlines extends Task {
```
yiwern5: Is it better to use a singular instead of plural noun for the class name? (same goes for ToDos and Events, considering Task is in singular)

src/main/java/Task.java
```diff
@@ -0,0 +1,32 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
```
yiwern5: Good job in correctly naming the boolean

src/main/java/Task.java
```diff
@@ -0,0 +1,32 @@
...
+
+    public void mark() {
+        this.isDone = true;
+    }
+
+    public boolean isDone() {
+        return isDone;
+    }
+
+    public void unMark() {
```
yiwern5: Nice attempt at using camel case, but I believe 'unmark' should be treated as a single term

src/main/java/ToDos.java
```diff
@@ -0,0 +1,9 @@
+public class ToDos extends Task{
```
yiwern5: Task should be followed by a whitespace
```suggestion
public class ToDos extends Task {
```

src/main/java/Univus.java
```diff
@@ -0,0 +1,229 @@
+import java.io.*;
```
yiwern5: It would be preferable to avoid using wildcard imports

src/main/java/UnivusException.java
```diff
@@ -0,0 +1,5 @@
+public class UnivusException extends Throwable{
```
yiwern5: Throwable should be followed by a whitespace
```suggestion
public class UnivusException extends Throwable {
```

src/main/java/Events.java
```diff
@@ -0,0 +1,14 @@
+public class Events extends Task{
```
yiwern5: Task should be followed by a whitespace
```suggestion
public class Events extends Task {
```

src/main/java/Univus.java
```diff
@@ -0,0 +1,229 @@
...
+            System.out.println("____________________________________________________________");
+        } else if (message.matches("mark\\s\\d+")) {
+            try {
+                int index = Integer.parseInt(message.split(" ")[1]);
+
+                if (index > store.size() || index < 1) {
+                    throw new UnivusException("OOPS!!! Wrong Index!");
+                }
+                Task task = store.get(index - 1);
+                if (task.isDone()) {
```
yiwern5, line 45: I appreciate how you handled the situation where users should be unable to mark the task as done when it is already marked as done, same goes for unmark

yiwern5 (commented):
LGTM! Overall, your code is clean and easy to understand. The main class is quite long, maybe can consider making it more OOP on your subsequent increment! I've added some suggestions for you too 😊

src/main/java/Univus.java
yiwern5: I appreciate how you handled the situation where users should be unable to mark the task as done when it is already marked as done, same goes for unmark

tanjiajiajun (commented):
Hi, I am tasked to do a PR review on your code. I will be reviewing based on code quality guidelines

1. In your `Parser.java`, I see that you have repeated 
```
System.out.println("____________________________________________________________");
``` 
a lot, maybe you can consider creating a variable `private final String LINE`
2. In `Storage.java`, you have 
```
int endIndex = message.lastIndexOf("(");
int startIdx = message.indexOf(":");
```
It would be nice if you can standardise if you want to name them Index or Idx.

Other than that, the rest of the code looks good! LGTM.

PR #401, opened 2024-02-02
[Murugan Maniish] iP

Murugan-Maniish:
### Alastor

> "He-llo!" – Alastor ([Source](https://hazbinhotel.fandom.com/wiki/Alastor/Quotes))

Alastor frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn

All you need to do is,

1. download it from [here](https://github.com/Murugan-Maniish/ip/releases/tag/v0.1)
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you 😈 

And it is **FREE!**

Features:

- [x] Managing tasks
- [ ] Reminders (coming soon)

Alastor commands:
`list`
`mark` 
`unmark`
`todo`
`deadline` 
`event`
`find`
`bye`

```java
        new Alastor("./data/tasks.txt").run();
    }
```


src/main/java/Storage.java
```diff
@@ -0,0 +1,74 @@
...
+        try {
+            FileWriter writer = new FileWriter(this.filePath, true);
+            writer.write(task.toFile() + "\n");
+            writer.close();
+        } catch (IOException e) {
+            throw new DukeException("I'm afraid I've encountered an error while saving your tasks, my dear.\n");
+        }
+    }
+
+    public void saveRewrite(TaskList tasks) throws DukeException {
```
sebin0817: For all public methods (except getters/setters and overriden methods), you should add Javadoc comments for clarity!

src/main/java/AddCommand.java
```diff
@@ -0,0 +1,20 @@
+public class AddCommand extends Command {
```
jmsandiegoo, line 1: I like how you created classes for each related code logics such as for commands to ensure single responsibility principle. However, would it be better if the class would be in a package? I noticed it with the other classes as well.

src/main/java/Storage.java
jmsandiegoo: Javadocs would be great indeed! 👍

src/main/java/AddCommand.java
jmsandiegoo: I like how you created classes for each related code logics such as for commands to ensure single responsibility principle. However, would it be better if the class would be in a package? I noticed it with the other classes as well.

src/main/java/Command.java
```diff
@@ -0,0 +1,4 @@
+public abstract class Command {
+    public abstract void execute(TaskList tasks, Ui ui, Storage storage) throws DukeException;
+    public abstract boolean isExit();
```
jmsandiegoo: Could you consider adding one blank line between these two logical units?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,24 @@
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class Deadline extends Task {
```
jmsandiegoo: Aside from public methods, you can also try and include Javadoc header comments for your classes perhaps?

src/main/java/Parser.java
```diff
@@ -0,0 +1,87 @@
...
+                return new AddCommand(new Event(event[0].trim(), stringToDateTime(event[1].trim()),
+                        stringToDateTime(event[2].trim())));
+            case "delete":
+                checkForEmpty(task, 2);
+                return new DeleteCommand(stringToInt(task[1]));
+            case "bye":
+                return new ExitCommand();
+            default:
+                return new InvalidCommand();
+        }
```
jmsandiegoo: Should the case have the same indentation as the switch line for this statement?

jmsandiegoo (commented):
LGTM 💯, just some code style issues need to be fixed.

PR #417, opened 2024-02-03
[JerryWang0000] iP

JerryWang0000:
# Duke

> "Yes we're all using it." – Eric Cartman ([source](https://www.youtube.com/watch?v=6lMaepNvP6A&t=89s))

Duke is a simple tool for task management and saves your effort of remembering that much stuff. It can: 

- tell you what need to do
- check if something is already done
- delete stuff you don't care any more

You just need to
1. Extract the Jar file
2. Run it just like how you run every Jar file :rocket:

You can use it for **FREE**.

Features:

- [X] Manage your tasks
- [X] Search by dates
- [ ] Modifying the content of tasks(Coming soon)

If you are a Java Programmer, here is the 'main()' method for your reference:

```Java
        Application.launch(Duke.class, args);
    }

### CODE REVIEW

File: src/main/java/BotManager.java [hunk 1]
```diff
@@ -0,0 +1,150 @@
+import java.util.ArrayList;
+
+public class BotManager {
+    private final String name;
+    private ArrayList<Task> tasks;
```
itstrueitstrueitsrealitsreal: I would consider making tasks a final attribute since you shouldn't be reassigning it later on.
```suggestion
    private final ArrayList<Task> tasks;
```

src/main/java/BotManager.java
```diff
@@ -0,0 +1,150 @@
...
+
+    BotManager(String name) {
+        this.name = name;
+        this.tasks = new ArrayList<Task>();
+    }
+
+    void greeting() {
+        System.out.println("    Hello! I'm " + name);
```
itstrueitstrueitsrealitsreal: I like how you abstracted the name out so you can change it later on!

src/main/java/BotManager.java
```diff
@@ -0,0 +1,150 @@
...
+public class BotManager {
+    private final String name;
+    private ArrayList<Task> tasks;
+
+    BotManager(String name) {
+        this.name = name;
+        this.tasks = new ArrayList<Task>();
+    }
+
+    void greeting() {
```
itstrueitstrueitsrealitsreal: I would change the name of the function to be greet instead of greeting in line with the coding standard.
```suggestion
    void greet() {
```

src/main/java/BotManager.java
```diff
@@ -0,0 +1,150 @@
...
+
+    void exit() {
+        System.out.println("    Bye. Hope to see you again soon!\n");
+    }
+
+    private Task createTask(String prompt) throws DukeException {
+        String[] order = prompt.split(" ", 2);
+        String taskType = order[0];
+        switch (taskType) {
+        case "todo":
```
itstrueitstrueitsrealitsreal: Like that you adhered to the coding standard for something like switch case statements, a lot of my peers overlooked this.

src/main/java/Duke.java
```diff
@@ -1,10 +1,19 @@
...
+
 public class Duke {
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        BotManager manager = new BotManager("Jerry");
```
itstrueitstrueitsrealitsreal: Interesting decision to name the bot after yourself...

src/main/java/Task.java
```diff
@@ -0,0 +1,30 @@
+abstract public class Task {
+    private final String name;
+    private boolean done;
```
itstrueitstrueitsrealitsreal: I would suggest that done be changed to isDone to make it sound more like a boolean, according to the coding standard for this course.
```suggestion
    private boolean isDone;
```

itstrueitstrueitsrealitsreal (commented):
Other than a few nits, LGTM!

src/main/java/BotManager.java
headcube1: Nice try on using verbs for methods, however maybe changing to greet would have been better?

src/main/java/Task.java
headcube1: Changing to isDone would have been better to follow the standard? For other boolean variables as well!

src/main/java/Task.java
```diff
@@ -0,0 +1,30 @@
...
+
+    Task(String name) {
+        this.name = name;
+        this.done = false;
+    }
+
+    boolean isDone() {
```
headcube1: Nicely done in naming the boolean method as isDone()!

src/main/java/BotManager.java
```diff
@@ -0,0 +1,150 @@
+import java.util.ArrayList;
+
+public class BotManager {
```
headcube1: In general the names of the methods and variables follow the coding standard!

headcube1 (commented):
Nice work!

PR #428, opened 2024-02-04
[Koh Tee Sing Newton] iP

NewtonKoh:
# Jelly

> "You are the Jelly to my Peanut Butter" -Someone Probably ([source](https://google.com/))

Jelly is a chatbot who can help you manage tasks, events, and deadlines! He is:

- Fun and Encouraging
- Full of Energy
- ~~Easy~~ SUPER EASY to use

All you need to do is:

1. download it.
2. double-click it.
3. add your tasks.
4. let it manage your tasks for you. :+1:

And it is **FREE**

Features:

- [x] Manage tasks.
- [ ] ChatGPT support for conversations. (coming soon)
- [ ] Reminders. (coming in the future)

If you want to learn how to code, Here's the `main` method to print "Hi Jelly" to your console:

```java
public class Jelly {
        System.out.println("Hi Jelly");
    }
}
```


src/main/java/Deadline.java
```diff
@@ -0,0 +1,49 @@
...
+            this.isParsed = true;
+        } catch(DateTimeParseException e){
+
+            System.out.println("Date cannot be parsed, stored as String instead.");
+        }
+    }
+
+    @Override
+    public String toString(){
+        if(isParsed){
```
BryanL2303: There should be a space between 'if' and the open bracket '('.

In almost all places across all the classes you seem to have missed the space before the '(' when they are necessary. Such as while and for loops and if/else statements.

src/main/java/Deadline.java
BryanL2303, line 1: In every line where the class or methods are defined there is no space before '{'.
This also seems to be the case for the other classes.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,49 @@
...
+
```
BryanL2303: I don't think the coding standard mentions anything about leaving blank lines after the declaration of a method or class, but is there a reason why you do this? The toString method above does not have the blank line for example.

In all the other classes I also see many blank lines in random spots... Not sure if they are placed there on purpose but I can't really understand the logic of having those spaces at those spots.

src/main/java/Jelly.java
```diff
@@ -0,0 +1,183 @@
...
+    public static String Read(Scanner scanner){
+
+        String message = scanner.nextLine();
+        String[] lines = message.split("\\s+");
+        String command = lines[0];
+
+        String argument = "";
+
+        if(lines.length > 1) {
+        argument = message.substring(command.length() + 1);
```
BryanL2303: Might have missed an identation here.

src/main/java/Jelly.java
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        System.out.println(line);
+
+        switch(command){
```
BryanL2303: Refering to the coding standard for switch commands the identations are different from what we might be used to. The lines of 'case' statements should be on the same identation level as the 'switch' statement itself. Only the items within the cases needs to be further indented

src/main/java/TaskList.java
```diff
@@ -0,0 +1,202 @@
...
+        System.out.println("(ᗒᗣᗕ) OK, I've marked this task as not done yet:");
+        System.out.println(task);
+    }
+
+    @Override
+    public String toString(){
+
+        String string = "";
+
+        for(Integer i=1;i<=tasks.size();++i){
```
BryanL2303: Generally we would always leave a space between operators.
I think this line should look like "Integer i = 1; i <= tasks.size(); ++i)"

PR #434, opened 2024-02-04
Create a PR to the upstream repo

jeong-jaeho:
Duke frees your mind of having to remember things you need to do. It's,
* text-based
* easy to learn
* ~~FAST~~ _SUPER_ FAST to use

All you need to do is,
1. download it from [here](https://github.com/nus-cs2103-AY2324S2/ip.git).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:
- [ ] Managing tasks
- [ ] Reminders (coming soon)

```java
public class Main {
    public static void `main`(String[] args) {
    }
}
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,133 @@
+import java.util.*;
```
cheahTJ: Hi Jaeho!! I feel it would be good to import the specific classes you need from the java util class as it is a good practice to avoid using the wildcard (`*`) :)

src/main/java/Deadline.java
```diff
@@ -0,0 +1,14 @@
...
+
+    protected String by;
+
+    public Deadline(String description, String by) {
+        super(description);
+        this.by = by;
+    }
+
+    @Override
+    public String toString() {
```
cheahTJ: I like how you have overriden the `toString()` method to make it specific to the deadline class. Remember also to include in your javadoc for this method to provide a justification as to why you have overriden the `toString()` method 👍

src/main/java/Duke.java
```diff
@@ -1,10 +1,133 @@
...
+                    Integer indexOfTask = Integer.parseInt(input.substring(7));
+                    Task currTask = tasks.get(indexOfTask);
+                    currTask.markAsNotDone();
+                    System.out.println("OK, I've marked this task as not done yet:");
+                    System.out.println(currTask);
+                } else if (input.startsWith("todo")) {
+                    // Check if the input string is long enough
+                    if (input.length() <= 5) {
+                        throw new DukeException("Please specify the description of the todo task, " +
+                                "starting from one whitespace away from the keyword 'todo'" +
```
cheahTJ: Really good practice of the correct indentation here. It made string message so easy to read.

src/main/java/Duke.java
```diff
@@ -1,10 +1,133 @@
...
+                    }
+
+                    Task todo = new Todo(description);
+                    tasks.add(todo);
+                    System.out.println("Got it. I've added this task: \n" + todo);
+                    System.out.println("Now you have " + tasks.size() + " tasks in the list.");
+                } else if (input.startsWith("deadline")) {
+                    String[] substrings = input.split(" /by ");
+
+                    if (substrings[0].length() <= 9) {
```
cheahTJ: Really good break (empty space) between each logical unit. It makes it so easy to understand your code!

src/main/java/Duke.java
```diff
@@ -1,10 +1,133 @@
+import java.util.*;
+
```
cheahTJ: I really like how you have managed to have the appropriate standard spacing (e.g. before a '{') in your code. Continue the good effort!

src/main/java/Duke.java
jeong-jaeho: Thanks for the suggestion!

src/main/java/Task.java
```diff
@@ -0,0 +1,31 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
```
ondretann: I like how you named your boolean variable! :)

src/main/java/Duke.java
ondretann: I agree with this. Correct indentation for "if else" has helped with readability for the logic of the code too.

ondretann (commented):
LGTM

PR #440, opened 2024-02-04
[Kim Junseo] iP

jskimdev:
### ByteTalker

ByteTalker manages your time by remembering things you need to do. It's,

- text-based
- easy to use
- SUPER FAST

All you need to do is,

1. download it from here.
2. double-click it.
3. add your tasks.

And it is FREE!

Features:

- [x] Managing tasks
- [x] Managing deadlines
- [x] Managing events

```
public class Main {
    }
}
```


src/main/java/ByteTalker.java
```diff
@@ -0,0 +1,135 @@
...
+    private ArrayList<Task> storage = new ArrayList<>();
+
+    public void returnList () {
+        System.out.println("    Here are the tasks in your list:");
+        for (int i = 0; i < storage.size(); i++) {
+            System.out.println("    " + (i + 1) + "." + storage.get(i).toString());
+        }
+    }
+
+    public void markTask(String[] split_message) {
```
guanquann: In Java, I believe the standard naming convention is camelCase. `split_message` should be `splitMessage` instead.

I notice the same issue in several other places as well

src/main/java/ByteTalker.java
```diff
@@ -0,0 +1,135 @@
+import java.util.ArrayList;
+import java.util.Scanner;
+
+public class ByteTalker {
```
guanquann: Javadocs not present in class and methods. 

It should look something like this:
```
/**
 * Returns lateral location of the specified position.
 * If the position is unset, NaN is returned.
 *
 * @param x X coordinate of position.
 * @param y Y coordinate of position.
 * @param zone Zone of position.
 * @return Lateral location.
 * @throws IllegalArgumentException  If zone is <= 0.
 */
public double computeLocation(double x, double y, int zone)
        throws IllegalArgumentException {
    //...
}
```

src/main/java/ByteTalker.java
```diff
@@ -0,0 +1,135 @@
...
+    private ArrayList<Task> storage = new ArrayList<>();
+
+    public void returnList () {
```
guanquann:
```suggestion
    public void returnList() {
```

src/main/java/Task.java
```diff
@@ -0,0 +1,68 @@
...
+    DEADLINE("D") {
+        @Override
+        public String toString() {
+            return "[" + getIcon() + "]" + "[" + getStatusIcon() + "] " + getTask() + " (by: " + getTo() + ")";
+        }
+    },
+    Event("E") {
+        @Override
+        public String toString() {
+            return "[" + getIcon() + "]" + "[" + getStatusIcon() + "] " + getTask() + " (from: " + getFrom() + " to:" + getTo() + ")";
```
guanquann: Line length should be no longer than 120 chars. Consider using linebreak

src/main/java/Task.java
```diff
@@ -0,0 +1,68 @@
...
+    private boolean isDone;
+    private String icon;
+    private String to;
+    private String from;
+    Task(String icon) {
+        this.icon = icon;
+        this.isDone = false;
+    }
+
+    public void setTask(String task) {
```
guanquann: Good use of getters and setters method!

src/main/java/ByteTalker.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    public static void main(String[] args) {
+        /*
+        String logo = " ____        _        \n"
+                + "|  _ \\ _   _| | _____ \n"
+                + "| | | | | | | |/ / _ \\\n"
+                + "| |_| | |_| |   <  __/\n"
+                + "|____/ \\__,_|_|\\_\\___|\n";
+        */
```
guanquann: This comment seems unnecessary to me. Consider removing them.

src/main/java/ByteTalkerException.java
```diff
@@ -0,0 +1,19 @@
+public class ByteTalkerException {
+    public static class TODONoTaskException extends Exception {
```
guanquann: The naming convention should not be in uppercase. Only constants should have uppercase. 

```suggestion
    public static class TodoNoTaskException extends Exception {
```
Same issue for DEADLINEException

guanquann (commented):
It seems like you are a little behind schedule. You should have completed the taskings in Week 3 already. Most of the comments should be corrected if you had completed the taskings in week 3.

You can take a look at [conventions](https://se-education.org/guides/conventions/java/intermediate.html) for more details on good coding standards. All the best!

src/main/java/bytetalker/exception/ByteTalkerException.java
```diff
@@ -0,0 +1,39 @@
...
+        }
+    }
+
+    public static class DeadlineWrongFormatException extends Exception {
+        public DeadlineWrongFormatException(String errorMessage) {
+            super(errorMessage);
+        }
+    }
+
+    public static class DeadlineUnsupportedFormatException extends Exception {
```
billyhoce: Seems like there is a `DeadlineWrongFormatException` and a `DeadlineUnsupportedFormatException`, could you share the reason to separate the two? It sounds like either can be used in the situation of a different format than expected!

src/main/java/bytetalker/parser/Parser.java
```diff
@@ -0,0 +1,104 @@
...
+            tempMessage += splitMessages[i] + " ";
+        }
+        ArrayList<String> resultMessages = new ArrayList<>();
+        tempMessage = tempMessage.strip();
+        resultMessages.add(tempMessage);
+        return resultMessages;
+    }
+
+    public static ArrayList<String> parseDeadlineInput(String[] splitMessages) throws ByteTalkerException.DeadlineUnsupportedFormatException {
+        ArrayList<String> storeMessages = new ArrayList<>();
```
billyhoce: An ArrayList object here is named with a verb name `storeMessages`, maybe a more noun-like name like `messageContainer` or `messageStore` can be considered!

src/main/java/bytetalker/task/TaskType.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @author Junseo Kim
+ * @version 1.0
+ * @since 2024-02-06
+ */
+public enum TaskType {
+    TODO("T") {
+    },
+    DEADLINE("D") {
+
```
billyhoce: Maybe you can standardise the formatting for the enum here. The first object "TODO" has no empty line after the curly bracket but "DEADLINE" and "EVENT" has one.

src/main/java/bytetalker/ui/Ui.java
```diff
@@ -0,0 +1,144 @@
...
+    public void showBye() {
+        System.out.println("    -----------------------------------");
+        System.out.println("    Bye. Hope to see you again soon!");
+        System.out.println("    -----------------------------------");
+    }
+
+    /**
+     * Prints out error message if there is an error while changing the task list.
+     */
+    public void showErrorMsgStoreTasks() {
```
billyhoce: Maybe an alternate name to consider would be "showStoreTaskErrorMessage"

billyhoce (commented):
Overall well written code that is neat and easy to follow! Just a few very small fixes. Keep it up!

src/main/java/ByteTalker.java
jskimdev: Ah missed that. Noted. Thanks for informing it.

src/main/java/ByteTalker.java
jskimdev: Added last week after the feedback!

src/main/java/Task.java
jskimdev: Noted, I will try to break the lines exceeding 120 chars

src/main/java/Task.java
jskimdev: Thanks!

src/main/java/bytetalker/exception/ByteTalkerException.java
jskimdev: At the time of coding it, I was thinking of different errors that could rise, but realised that both errors are eventually the same! I will try to combine both exceptions

src/main/java/bytetalker/parser/Parser.java
jskimdev: Thank you for feedback! I will think about it

src/main/java/bytetalker/task/TaskType.java
jskimdev: Oh, alright I will try to standardise the format. Thank you for feedback!

src/main/java/bytetalker/ui/Ui.java
jskimdev: That seems to make more sense. Thank you for feedback!

PR #442, opened 2024-02-04
[matthewken19] iP

matthewken19:
DukePro frees your mind of having to remember things you need to do. It's,
- text-based
- easy to learn
- ~~FAST~~SUPER FAST to use

All you need to do is,
1. download it from [here](https://www.google.com/).
2. double-click it
3. add your tasks
4. let it manage your tasks for you 👍 

And it is **FREE**!

Features:
- [x] Managing tasks
- [x] GUI
- [ ] Find deadlines (coming soon)
--- 
If you are a Java Programmer, you can use it to practice Java too. Here's the `main` method:
```
public class Main {
    }
}
```


src/main/java/Duke.java
```diff
@@ -1,10 +1,267 @@
...
 public class Duke {
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+
+        File dir;
```
AndrewOng2066:
```suggestion
        File directory;
```
Perhaps the name can be spelt out fully in case someone might not understand what "dir" stands for

src/main/java/Duke.java
```diff
@@ -1,10 +1,267 @@
...
+        File f;
```
AndrewOng2066:
```suggestion
        File file;
```
Using "file" instead of "f" as the variable name would be preferable, especially since the variable is referenced elsewhere in the code. It can be challenging for people to recall the significance of "f" later on.

src/main/java/Duke.java
```diff
@@ -1,10 +1,267 @@
...
+        ArrayList<Task> list = new ArrayList<>();
```
AndrewOng2066:
```suggestion
        ArrayList<Task> taskList = new ArrayList<>();
```
Can consider changing to a more meaningful name instead of "lists" as it can refer to any list instead of a specific list of tasks.

src/main/java/Duke.java
```diff
@@ -1,10 +1,267 @@
...
+
+        String line = "_____________________________________________";
+        String greeting = "Hello! I'm Donald.\nWhat can I do for you?\n";
+        System.out.println(line);
+        System.out.println(greeting + line + "\n");
+
+        Scanner sc = new Scanner(System.in);
+        String input;
+        int index;
+        Task t;
```
AndrewOng2066:
```suggestion
        Task task;
```
Perhaps the name can be spelt out fully, same as the above.

src/main/java/Duke.java
```diff
@@ -1,10 +1,267 @@
...
+        } catch (IOException e) {
+            System.out.println("Something went wrong.. Failed to save list.");
+        }
+
+        String closing = "Bye. Hope to see you again soon!\n";
+        System.out.println(closing + line);
+    }
+
+    private static ArrayList<Task> readFileContents(File filePath) throws FileNotFoundException {
+        ArrayList<Task> list = new ArrayList<>();
```
AndrewOng2066:
```suggestion
        ArrayList<Task> tasklist = new ArrayList<>();
```
Can consider changing to a more meaningful name instead of "lists" as it can refer to any list instead of a specific list of tasks.

src/main/java/Duke.java
```diff
@@ -1,10 +1,267 @@
...
+        Scanner sc = new Scanner(filePath);
+        String s;
```
AndrewOng2066: Perhaps can use a more meaningful name for this variable.

src/main/java/Duke.java
```diff
@@ -1,10 +1,267 @@
...
+        Task t;
+        while (sc.hasNext()) {
+            s = sc.nextLine();
+            try {
+                stringSplit = s.split("-");
+                for (int i = 0; i < stringSplit.length; i++) {
+                    stringSplit[i] = stringSplit[i].trim();
+                    System.out.println(stringSplit[i]);
+                }
+                boolean marked = Boolean.valueOf(stringSplit[1]);
```
AndrewOng2066:
```suggestion
                boolean isMark= Boolean.valueOf(stringSplit[1]);
```
Can consider changing the name of the boolean variable to sound like booleans.

src/main/java/Duke.java
```diff
@@ -1,10 +1,267 @@
...
+                switch (stringSplit[0]){
```
AndrewOng2066:
```suggestion
                switch (stringSplit[0]) {
```
Missing a space before the curly bracket.

AndrewOng2066 (commented):
Overall, everthing is okay except some variable names might need to change.

src/main/java/duke/AddCommand.java
```diff
@@ -0,0 +1,58 @@
+package duke;
+
+public class AddCommand implements Command {
+
+    private String commandWord;
+    private String description;
+    static String defaultString = "Got it. I've added this task:";
```
Zack-Tay: I believe if this variable here is a constant it can be in all caps.

src/main/java/duke/AddCommand.java
```diff
@@ -0,0 +1,58 @@
...
+
+    AddCommand(String commandWord, String description) {
+        this.commandWord = commandWord;
+        this.description = description;
+    }
+
+    public void execute(TaskList tasks, Ui ui) throws DukeException {
+        Task task;
+        String[] stringSplit;
```
Zack-Tay: for naming convention sake maybe you want to consider changing it to something like stringComponents?

src/main/java/duke/Ui.java
```diff
@@ -0,0 +1,54 @@
+package duke;
+
+import java.util.Scanner;
+
+public class Ui {
+    static String LINE = "_____________________________________________";
```
Zack-Tay: Good use of all CAPS for constants

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,65 @@
+package duke;
+
```
Zack-Tay: Might want to start adding javaDocs early to lesser workload in the coming weeks!

src/main/java/duke/DukeException.java
```diff
@@ -0,0 +1,8 @@
+package duke;
+
+public class DukeException extends Exception {
```
Zack-Tay: If your bot name is Donald then ideally your Exception should be DonaldException rather than DukeException, same goes for the main class name!

Zack-Tay (commented):
Overall, I like the OOP abstraction! LGTM, but just take note of some changes that I have proposed that might help!

PR #445, opened 2024-02-04
[laney0808] iP

laney0808:
### Rick Chatbot

Rick helps you to keep track of to-dos, events and deadlines.

- add _to do items_
- add deadlines and events with date and time
- mark items as done or unmark items as undone

Rick helps you achieve **seamless** workflow.

### To use Rick ✨

1. Download [Rick Latest Version](https://github.com/laney0808/ip/releases/tag/A-Jar)
2. Wait for the initialization
3. Add your tasks by inputing command "todo [your todo item]"
4. Let Rick do the rest!
5. ~Try out erratic commands to have a taste of Rick's error messages~

- [ ] I agree to Terms and Conditions in using Rick

> "Rick helps me to remember all the things that I do not bother to." - Amily, Rick's precious client

In case you need the `main` function, it is here:

```java
public class Main {
    }
}
```


src/main/java/Deadline.java
```diff
@@ -0,0 +1,25 @@
...
+        }
+        if (ddl.isBlank()) {
+            throw new RickException("due when?");
+        }
+        this.name = name;
+        this.ddl = ddl;
+    }
+    public String print(){
+        return "[D]" + this.status + this.name + " (by: " + ddl + ")";
+    };
```
alfaloo: why does this method have a semicolon after the curly bracket

src/main/java/Rick.java
```diff
@@ -0,0 +1,134 @@
...
+                "What can I do for you ?\n";
+        reply(hello);
+    }
+
+    public static void reply(String arg) {
+        String divider = "\n____________________________________________________________\n";
+        System.out.println(divider + arg + divider);
+    }
+
+    public static void exit(){
```
alfaloo: do you think it would be better to keep curly bracket placement consistent (other methods have a white space after the round bracket and before the curly bracket)

src/main/java/RickException.java
```diff
@@ -0,0 +1,10 @@
+public class RickException extends Exception{
```
alfaloo: do you think you should keep consistency with spacing? the other exceptions have a white space before the curly bracket

src/main/java/ToDo.java
```diff
@@ -0,0 +1,20 @@
...
+    public String print() {
+        return "[T]" + this.status + this.name;
+    }
+    public void mark() {
+        this.status = "[X]";
+    }
+    public void unmark() {
+        this.status = "[ ]";
+    }
+}
```
alfaloo: do you think it would be better implementing an overriding toString method to display the tasks?

src/main/java/Deadline.java
alfaloo: do you think its a cleaner solution to add an overriding toString method?

alfaloo (commented):
Overall very well done. Just some nitpicks. Keep up with the good work.

PR #449, opened 2024-02-05
 [jayllo-c] iP

jayllo-c:
# Destiny
> “Your mind is for having ideas, not holding them.” – David Allen ([source](https://dansilvestre.com/productivity-quotes))
Destiny frees your mind of having to remember things you need to do. It's,
- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use
All you need to do is,
1. download it from [here](https://github.com/jayllo-c/ip/releases/tag/A-UserGuide).
2. double-click it.
3. add your tasks.
And it is **FREE**
Features:

- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon)
##
```
public class Main {
    }
}
```


src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        try {
+            this.by = LocalDateTime.parse(time, inputFormatter);
+        } catch (DateTimeParseException e) {
+            throw new DukeException("Please enter the date and time in the following format:\n"
+                    + "dd/mm/yyyy hhmm (e.g. 30/01/2024 1234");
+        }
+
+        switch (logic) {
```
kaitinghh: Since you only have one case, would an if else be more concise than switch?

src/main/java/duke/DukeException.java
```diff
@@ -0,0 +1,14 @@
+package duke;
+
+/**
+ * An extension of the Exception class used to identify exceptions unique to Duke.
+ */
+public class DukeException extends Exception {
```
kaitinghh: Could there be a better naming for DukeException class?

src/main/java/duke/Storage.java
```diff
@@ -0,0 +1,118 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Scanner;
+
+/**
+ * Used to load/store TaskList from/to a duke.txt data file.
+ */
+public class Storage {
+
+    private final String dataPath = System.getProperty("user.dir") + "/data";
```
kaitinghh: Final fields should be all capsed?

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,93 @@
...
+
+import java.util.ArrayList;
+
+/**
+ * Used to store and manage the list of tasks.
+ */
+public class TaskList {
+    private String horzLine = "____________________________________________________________";
+
+    private ArrayList<Task> taskList = new ArrayList<>(100);
```
kaitinghh: Why do you have to initialise the size of the taskList if its an arrayList?

kaitinghh (commented):
LGTM! Some minor comments for your review.

saiutkarsh33:
LGTM!

src/main/java/duke/Duke.java
1rbg, line 1: The Ui class should handle all UI-related functions, which include the printing of messages and the formatting to the terminal. Should all these functionality be moved to the Ui class?

src/main/java/duke/ToDo.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /**
+     * Secondary constructor used by Duke to generate a ToDo object from a data txt file.
+     *
+     * @param logic Indicates if this task should be marked.
+     * @param description Title of the task.
+     * @throws DukeException If description is empty.
+     */
+    public ToDo(String logic, String description) throws DukeException {
```
1rbg: Using `logic` as a variable name might not be very clear as it is not apparent what it is, but naming it with an 'is_' prefix is also not right as it isn't exactly a boolean, maybe you could rename this to `marked` or something similar, just a suggestion.

src/main/java/duke/ToDo.java
```diff
@@ -0,0 +1,46 @@
...
+            throw new DukeException("Please enter a description for this todo task");
+        }
+        switch (logic) {
+        case "1":
+            this.isDone = true;
+            break;
+        default:
+            this.isDone = false;
+            break;
+        }
```
1rbg: If there are only 2 cases to be handled here, "1", or everything else, could using an if statement be better?

src/main/java/duke/Storage.java
1rbg: final and constant are not the same thing though, typically constants refer to class members, final just means it cannot be modified after instantiation. I think it's fine that they are not all capitalised since they are instance fields, however perhaps if the variables are class members, and then they should be in all caps.

1rbg (commented):
I like that your code is pretty well organised, except for maybe the Ui functionality still residing in Duke. I have some minor comments about the code quality, and provided some suggestions. Overall well done so far!

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,91 @@
...
+        super(description);
+        time = time.trim();
+        if (description.trim().length() == 0 || time.length() == 0) {
+            String errorMessage = "Please enter the following:";
+            if (description.trim().length() == 0) {
+                errorMessage += "\na description for this duke.Deadline task";
+            }
+            if (time.length() == 0) {
+                errorMessage += "\na deadline after the '/by' command";
+            }
```
LWS49: That's a good way to customize your error message!

src/main/java/duke/DukeException.java
LWS49: Could it be better to have different types of DukeException instead? With this way of implementation all error messages would be in this file, and can be edited without going into other files unnecessarily. (encapsulation)

src/main/java/duke/TaskList.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+        if (index < 1 || index > taskList.size()) {
+            throw new DukeException("Please enter a number between 1 and " + taskList.size());
+        }
+        Task tempTask = taskList.get(index - 1);
+        taskList.remove(index - 1);
+        printWithLines("Noted. I've removed this task:\n   " + tempTask.toString()
+                + "\nNow you have " + taskList.size() + (taskList.size() > 1 ? " tasks " : " task ")
+                + "in the list.");
+    }
```
LWS49: this should be under UI, right? Since UI should be handling all outputs.

PR #457, opened 2024-02-05
[beefsausagee] iP

alvislowyh:
# XiaoBai
>“Your mind is for having ideas, not holding them.” – David Allen ([source](https://dansilvestre.com/productivity-quotes))

XiaoBai frees your mind of having to remember things you need to do. It's,

- text-based
- easy to learn
- ~~FAST~~ SUPER FAST to use

All you need to do is,

1. download it from [here](https://github.com/beefsausagee/ip/releases/tag/A-Release).
2. double-click it.
3. add your tasks.

And it is **FREE**!

Features:

- [x]  Managing tasks

```
public class Main {
    }
}
```


src/main/java/Duke.java
```diff
@@ -1,8 +1,51 @@
...
+                bye();
+                hasEnded = true;
+            } else {
+                repeatFunction(input);
+            }
+        }
+    }
+
+    /**
+     * Displays a greeting message
```
kwuunnn: I think the sentence should end with a fullstop. I noticed the same issue in several other places too.

src/main/java/Duke.java
```diff
@@ -1,8 +1,51 @@
...
+     * 
+     * @param botName Name of the bot
```
kwuunnn: Likewise, I think each parameter description should end with a fullstop. I noticed the same issue in several other places too.

src/main/java/Duke.java
```diff
@@ -1,8 +1,51 @@
...
+     */
+    public static void greet(String botName) {
```
kwuunnn, line 37: Not sure there's a need to make the botName a parameter since it's probably only going to be used during greeting. Maybe just hardcode the botName in the greeting message? Then there would be no need to declare the botName variable in main().

src/main/java/Duke.java
```diff
@@ -1,8 +1,51 @@
...
 public class Duke {
-    public static void main(String[] args) {
-        String logo = "Hello! I'm Yube \n"
-                + "What can I do for you? \n" +
-                "Bye. Hope to see you again soon!";
-        System.out.println(logo);
+    private static final String LINE = "___________________________________________________________\n";
+
+    public static void main(String[] args) throws IOException {
+        BufferedReader reader = new BufferedReader(new InputStreamReader(System.in));
```
kwuunnn, line 10: Interesting how you used BufferedReader instead of Scanner. Is there any advantages of using it?

src/main/java/Duke.java
```diff
@@ -1,8 +1,51 @@
...
+        System.out.println(String.format(
+                "%sBye. Hope to see you again soon! \n%s", LINE, LINE));
+    }
+
+    /**
+     * Displays a repeated message of the input by the user
+     * 
+     * @param input User input
+     */
+    public static void repeatFunction(String input) {
```
kwuunnn, line 55: Not sure I like the name 'repeat'. Maybe echo would have been a more appropriate name?

kwuunnn (commented):
LGTM. Just a few nits to fix.

src/main/java/tempCodeRunnerFile.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /**
+     * Displays a repeated message of the input by the user and number of task in
+     * list.
+     * 
+     * @param task        Task input
+     * @param listOfTasks List of all tasks
+     */
+    public static void repeatFunction(Task task, ArrayList<? extends Task> listOfTasks) {
```
xuelinglow: Perhaps it would be better to use a more descriptive function name to make it more clear what the function does 
e.g notifyTaskAdded

xuelinglow (commented):
So far looks pretty good in terms of code quality guidelines.

minghowC (commented):
Well implemented. LGTM, just one nit to fix. Perhaps magic number could be better explained with named constants in the case of line 32? 
Pardon my PR at this juncture of time.

PR #461, opened 2024-02-05
[jasperng] iP

jasperng-nus:
# Luke
> “Learn from yesterday, live for today, hope for tomorrow. The important thing is not to stop questioning.” – Albert Einstein ([source](https://www.brainyquote.com/quotes/albert_einstein_125368))
- text based
- easy to use
- ~~SUPER~~ *FAST* to use

1. download from here.
2. double click it.
3. add your tasks.
4. let it manage yours tasks for you :+1:

And it is **FREE**!

Features:
- [x] managing task
- [x] add task
- [ ] delete task

> [!NOTE]
> There some things to note.

this is an inline `code` it has back-ticks `around` it

```java
public class Main {
        System.out.println("Hello World!");
    }
}

```


src/main/java/Deadline.java
```diff
@@ -0,0 +1,13 @@
+public class Deadline extends Task {
+    protected String by;
```
Austintjh19: Is using protected the most suitable option ? It makes it accessible to other classes from the same package ... I noticed the same issue in several other places too.

src/main/java/Event.java
```diff
@@ -0,0 +1,16 @@
+public class Event extends Task{
```
Austintjh19: Whitespace should proceed { , according to our coding standard. Just something to take note.

src/main/java/Luke.java
```diff
@@ -0,0 +1,205 @@
...
+        System.out.println("________________________________________________________________________");
+
+        // User inputs
+        Scanner sc = new Scanner(System.in);
+        String input = sc.nextLine();
+        ArrayList<Task> list = new ArrayList<>();
+
+        // Conditions
+        while (!input.equals("bye")) {
+            if (input.equals("list")) {
```
Austintjh19: Maybe consider creating functions to execute the commands ? E.g. when input.equals("list ") it call a separate function like outputTasks(). You can possibly perform you validation checks in these functions as well.

src/main/java/Luke.java
```diff
@@ -0,0 +1,205 @@
+import java.util.ArrayList;
+import java.util.Scanner;
+
+public class Luke {
```
Austintjh19: Maybe enhance the class structure by decomposing the functionality into separate classes such as InputValidator for validating user inputs, Outputs for managing output operations, and InputInterpreter for processing user commands, thereby promoting a clearer separation of concerns.

src/main/java/Luke.java
Austintjh19: Consider switching to a case statement it can potentially enhance code readability.

src/main/java/Luke.java
```diff
@@ -0,0 +1,205 @@
...
+
+                        if (!markOrUnmark.equals("mark") && !markOrUnmark.equals("unmark")) {
+                            System.out.println("test");
+                            throw new LukeException("Hold up!! I am sorry, but I don't know what you mean by that :'(");
+                        }
+
+                        int index = Integer.parseInt(instruction[1]) - 1;  // array is 0-indexed
+
+                        if (index >= list.size()) {
+                            throw new LukeException("Hold up!! There is no such task in the list.\n"
```
Austintjh19: Nice to see you are already incorporating Exceptions.

src/main/java/TaskException.java
```diff
@@ -0,0 +1,5 @@
+public class TaskException extends Exception {
```
Austintjh19: Maybe have this extend LukeException ? e.g. LukeTaskException. Won't have to check for LukeException | TaskException for each relevant exception.

src/main/java/Luke.java
```diff
@@ -0,0 +1,205 @@
...
+
+                for (int i = 0; i < list.size(); i++) {
+                    if (list.get(i) != null) {
+                        System.out.println(i + 1 + "." + list.get(i));
+
+                    }
+                }
+                System.out.println("________________________________________________________________________");
+
+            } else if (input.contains("mark")) {
```
Austintjh19: Might be a redundant if check. You can just split the string and verify if the mark/ unmark is present, while also ensuring the length of the split instruction is 2. This way, you confirm the task name's presence without extra checks. This might be applicable for other commands as well.

src/main/java/Luke.java
```diff
@@ -0,0 +1,205 @@
...
+        // Greetings
+        System.out.println("________________________________________________________________________");
+        System.out.println("Hello! I'm Luke");
+        System.out.println("What can I do for you?");
+        System.out.println("________________________________________________________________________");
+
+        // User inputs
+        Scanner sc = new Scanner(System.in);
+        String input = sc.nextLine();
+        ArrayList<Task> list = new ArrayList<>();
```
Austintjh19: Possibly a more appropriate variable name to make the code more readable.

src/main/java/Luke.java
```diff
@@ -0,0 +1,205 @@
...
+
+                } catch (LukeException e) {
+                    System.out.println("________________________________________________________________________");
+                    System.out.println(e.getMessage());
+                    System.out.println("________________________________________________________________________");
+                }
+
+            } else {
+                try {
+                    if (input.equals("todo") || input.equals("deadline") || input.equals("event")) {
```
Austintjh19: Maybe consider the case where user enters "todo " or "deadline " or "event ".

Austintjh19 (commented):
Good start. The basics are there just some modifications to the Luke class are needed.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,13 @@
+public class Deadline extends Task {
```
Austintjh19: JavaDoc comments of classes and functions would be nice.

src/main/java/Event.java
```diff
@@ -0,0 +1,16 @@
...
+    public Event(String description, String from, String to) {
+        super(description);
+        this.from = from;
+        this.to = to;
+    }
+
+    @Override
+    public String toString() {
+        return "[E]" + super.toString() + " (from: " + this.from + " to: " + this.to + ")";
+
```
Austintjh19: Trailing whitespaces are quite prevalent in your code. Just remove them and you're good.

Austintjh19 (commented):
Great work on coding standards. Seems like it was well followed.

src/main/java/luke/parser/Parser.java
```diff
@@ -0,0 +1,215 @@
+package luke.parser;
+
+import luke.Storage;
+import luke.exception.DateException;
+import luke.exception.FileException;
+import luke.exception.LukeException;
+import luke.exception.TaskException;
+import luke.task.*;
```
ChillinRage: Would it be better if you split the wildcard import and instead import each module explicitly?

src/main/java/luke/parser/Parser.java
```diff
@@ -0,0 +1,215 @@
...
+
+    /**
+     * Parses the user input and executes the corresponding command to add, delete, mark, unmark or list the tasks
+     * as well as to exit the program.
+     *
+     * @param list The list of tasks.
+     * @param ui The user interface.
+     * @param storage The storage of the tasks.
+     */
+    public void parse(TaskList list, Ui ui, Storage storage) {
```
ChillinRage: Will a switch-case work better instead of if-else blocks?

src/main/java/luke/parser/Parser.java
```diff
@@ -0,0 +1,215 @@
...
+                            + "Please enter an index after " + input + ".");
+                }
+
+                try {
+                    String[] instruction = input.split(" ");
+                    String delete = instruction[0];
+
+                    if (!delete.equals("delete")) {
+                        throw new LukeException("Hold up!! I am sorry, but I don't know what you mean by that :'(");
+                    }
```
ChillinRage: How about using ".startsWith("delete")" instead to check for commands? That way, you wouldn't need a separate if equals block.

src/main/java/luke/parser/Parser.java
```diff
@@ -0,0 +1,215 @@
...
+
+                } else if (type.equals("event")) {
+                    try {
+                        // 6 is the index after "event ", so starts from index 6
+                        // -1 to remove the space before "/from"
+                        String description = input.substring(6, input.indexOf("/from") - 1);
+                        // +6 to remove the "/from " and start from the index after "/from "
+                        // -1 to remove the space before "/to"
+                        String from = input.substring(input.indexOf("/from") + 6, input.indexOf("/to") - 1);
+                        // +4 to remove the "/to " and start from the index after "/to "
```
ChillinRage: I like that you use comments to explain the purpose of your code.

src/main/java/luke/task/Event.java
```diff
@@ -0,0 +1,59 @@
+package luke.task;
+
+import luke.exception.DateException;
+import luke.parser.DateTimeParser;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
```
ChillinRage, line 7: I like that you separated the imports by packages.

src/main/java/luke/ui/Ui.java
ChillinRage, line 1: Would it be better to have a separate function of constant variable to print the horizontal lines instead?

PR #464, opened 2024-02-06
[ananya21] iP

ananya21:
# Project Charlie Enhancement 🚀

> "Innovation distinguishes between a leader and a follower." - Steve Jobs

This PR introduces several enhancements to the **Project Charlie**, improving its performance and user experience. Here's what's new:

## Features:

- **Enhanced Speed**:
  - Now **_50% faster_** than before!
- **New User Interface**:
  - A sleek, modern UI for an improved user experience.

## How to Get Started:

1. **Download** the latest version from [here](https://google.com).
2. **Install** it by double-clicking the downloaded file.
3. **Enjoy** the new and improved features!

## What's Included in This PR:

```java
public class Enhancements {
        System.out.println("New Features Implemented!");
    }
}

### CODE REVIEW

File: src/main/java/Charlie.java [hunk 1]
```diff
@@ -0,0 +1,59 @@
+import java.util.ArrayList;
+import java.util.Scanner;
+
```
wilsonwid: It would be good to add a Javadoc comment for the class, as well as the public methods inside the class.

src/main/java/Charlie.java
```diff
@@ -0,0 +1,59 @@
+import java.util.ArrayList;
```
wilsonwid: It would be good to place all files in a package.

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
+
+    public Task(String description) {
```
wilsonwid: It would be good to add Javadoc comments to public methods and to the class.

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
...
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "[X]" : "[ ]"); // mark done task with X
+    }
+
+    public void markAsDone() {
```
wilsonwid: Could this method and `markAsNotDone()` perhaps be fused into one method with a `boolean` parameter?

src/main/java/Todo.java
```diff
@@ -0,0 +1,11 @@
+import java.util.ArrayList;
+
+public class Todo extends Task {
+    public Todo(String description) {
+        super(description);
+    }
+    @Override
+    public String toString() {
+        return "[T]" + super.toString();
```
wilsonwid: I like how inheritance and polymorphism are used to shorten this line.

wilsonwid (commented):
The code in this PR is of an overall good standard, even though there are several places where it could be improved. I commented on several coding standard violations, most of which are due to missing Javadoc comments. It would also be good to review the imports used in each of the classes to see whether they are unused.

NusMinato (commented):
No code naming issues found. Well done!

PR #469, opened 2024-02-06
Zhou Yang ip

SamuelZero1:
SamuelBot Pro
“Your mind is for having ideas, not holding them.” – David Allen

SamuelBot Pro frees your mind of having to remember things you need to do. It's,

text-based
easy to learn
FAST SUPER FAST to use
All you need to do is,
download it from [here](https://nus-cs2103-ay2324s2.github.io/website/admin/ip-w4.html).
double-click it.
add your tasks.
let it manage your tasks for you 😉
And it is FREE!

Features:

 Managing tasks
 Managing deadlines 
 Managing events

 Reminders (coming soon)

public class Main {
        Application.launch(ChatBotGUI.class, args);
    }
}


src/main/java/Duke_Level4.java
```diff
@@ -0,0 +1,85 @@
...
+public class Duke_Level4 {
+    public static void main(String[] args) {
+        String logo = " ____        _        \n"
+                + "|  _ \\ _   _| | _____ \n"
+                + "| | | | | | | |/ / _ \\\n"
+                + "| |_| | |_| |   <  __/\n"
+                + "|____/ \\__,_|_|\\_\\___|\n";
+        System.out.println("Hello from\n" + logo);
+        System.out.println("What can I do for you?");
+        List<String> list = new ArrayList<>();
```
ReflectiveObsidian, line 13: Could this variable name be updated to show that there are multiple objects represented? 
```suggestion
        List<String> tasks = new ArrayList<>();
```

ReflectiveObsidian (commented):
Just one nitpick.

src/main/java/Duke_Level4.java
```diff
@@ -0,0 +1,85 @@
...
+        String[] task = new String[100];
+        for (int i = 0; i < task.length; i++) {
+            task[i] = "[]";
+        }
+        String[] done = new String[100];
```
KohGuanZeh, line 18: Maybe this variable can use a better name such as `taskPrefixes`?

src/main/java/Duke_Level4.java
```diff
@@ -0,0 +1,85 @@
...
+        for (int i = 0; i < done.length; i++) {
+            done[i] = "[] ";
+        }
+        Scanner s = new Scanner(System.in);
+        String str = s.nextLine();
+        while(! str.equals("bye")){
+            if(str.substring(0, 4).equals("mark")){
```
KohGuanZeh, line 25: I think there should be a space in between `if`, `while` and the brackets?

src/main/java/Duke_Level1.java
```diff
@@ -0,0 +1,14 @@
+import java.util.Scanner;
+public class Duke_Level1 {
+    public static void main(String[] args) {
+        System.out.println("Hello! I'm SamuelBot");
+        System.out.println("What can I do for you?");
+        Scanner s = new Scanner(System.in);
+        String str = s.nextLine();
+        while(! str.equals("bye")) {
```
KohGuanZeh, line 8: I think there should not be a space between `!` and `str.equals("bye")`?

src/main/java/Duke_Level4.java
```diff
@@ -0,0 +1,85 @@
+import java.util.ArrayList;
```
KohGuanZeh, line 1: Perhaps consider having a package for `Duke`?

KohGuanZeh (commented):
I see that you have been making progress with the program as you are starting to incorporate more classes to better manage your programs. I think that the program can be further improved by using names more representative of what the variable is.

PR #476, opened 2024-02-07
Alex Setyawan iP

alex-setyawan:
### AWEX

> "When in doubt, look intelligent." - Garrison Keillor

And that is exactly what using AWEX allows you to do! Why record tasks using a notepad or a calendar, when you can painstakingly download a rudimentary project from online, that provides the exact same functionality?

Notepad? Boring. Overused. Plain.

Chatbot? Cool. Fancy. [_Sexy_.](https://en.wikipedia.org/wiki/Sexy)

AWEX (**A** **W**onderful **EX**perience) revolutionises the way we construct our to-do lists, being

- text-based
- easy to learn
- entertaining to watch,

and letting you,

- [x] create tasks
- [x] delete tasks
- [x] find tasks

unlike ANY other productivity app! 😱 

All you need to do is:

1. Wait for me to finish it.
2. Explore various better alternatives in the meantime.

Want to see how we make our tasks? Here's a sneak peek of our `Task` class!

```
public class Task {
    private String what;
    private String status;
    private String type;
    .
    .
    .
}

```


alvislowyh:
Add code descriptions and change method names such as.. message() to displayMessage?

src/main/java/Awex.java
```diff
@@ -0,0 +1,130 @@
+import java.util.*;
+
+public class Awex {
+    public static void message() {
```
Jajared: Good abstraction of the message to display the list of commands available. Perhaps a better and more specific variable name could be used here to describe this function? For example you could rename it to: `listAllCommands()`

src/main/java/Awex.java
```diff
@@ -0,0 +1,130 @@
...
+        System.out.println("  3. unmark <task number>");
+        System.out.println("  4. todo <task>");
+        System.out.println("  5. deadline <task> /by <deadline>");
+        System.out.println("  6. event <task> /from <start> /to <end>");
+        System.out.println("Type 'bye' to exit.");
+    }
+
+    public static void main(String[] args) {
+        System.out.println("Hello! I'm AWEX!\nWhat can I do for you?");
+        LinkedList<Task> list = new LinkedList<>();
```
Jajared: Any reason why a LinkedList is used instead of an ArrayList? Indeed a LinkedList works similar to an ArrayList, although their use cases may vary depending on the scenario. Here is some research that I did: [ArrayList vs LinkedList](https://www.w3schools.com/java/java_linkedlist.asp)

src/main/java/Awex.java
```diff
@@ -0,0 +1,130 @@
+import java.util.*;
```
Jajared: Perhaps you could try to avoid the use of wildcard imports here as it could lead to import conflicts.

src/main/java/Awex.java
```diff
@@ -0,0 +1,130 @@
...
+        System.out.println("Type 'bye' to exit.");
+    }
+
+    public static void main(String[] args) {
+        System.out.println("Hello! I'm AWEX!\nWhat can I do for you?");
+        LinkedList<Task> list = new LinkedList<>();
+        Scanner sc = new Scanner(System.in);
+        String next = sc.nextLine();
+        String[] arr = next.split(" ");
+        while (!next.equals("bye")) {
```
Jajared: Good use of if-else statements to check and parse inputs! However, perhaps you could abstract out this method and also do consider using switch-case statements to improve performance.

src/main/java/DeadlineTask.java
```diff
@@ -0,0 +1,12 @@
+public class DeadlineTask extends Task {
```
Jajared: Good use of inheritance!

Jajared (commented):
Overall good use of OOP principles and general adherence to coding standards. Some improvements and refactoring can be done to clean up the code and improve readability. Keep up the good work!

PR #477, opened 2024-02-07
[quelinxiao] ip

quelinxiao:
# Bob
Why `you` should use Bob:
> So you can work faster
* Text-based
* Easy to learn
* ~~FAST~~ SUPER FAST to use

Steps to take:
1. Download it [here](https://github.com/quelinxiao/ip)
2. Open the file
3. Start adding your tasks in :smile:

It is completely **FREE!**

Features:
- [x] Manage tasks
- [x] Manage deadlines
- [x] Find

This is the main method:
```
        new Bob("./data/tasks.txt").run();
}
```


src/main/java/Bob.java
```diff
@@ -31,7 +31,7 @@ else if (input.equals("list")) {
                 }
             }
 
-            else if (input.trim().matches("mark|unmark|deadline|todo|event")) {
+            else if (input.trim().matches("mark|unmark|deadline|todo|event|delete")) {
```
NewtonKoh: Small nit. I believe the else if should be preceded by the closing curly brackets from the if/if else statement above it.

NewtonKoh (commented):
LGTM. Small nits

src/main/java/Bob.java
```diff
@@ -0,0 +1,221 @@
+import java.util.Scanner;
+import java.util.ArrayList;
+
+public class Bob {
+    public static void main(String[] args) {
+        String greet = " Hello! I'm Bob.\n"
+                + " What can I do for you?\n";
+
+        String exit = " Bye. Hope to see you again soon!";
```
tanjiajiajun: Consider changing this to a final String as the value of the greeting and exit phrase will not change

src/main/java/Bob.java
```diff
@@ -0,0 +1,221 @@
...
+                String deadlineDate = input.substring(byIndex + 4).trim();
+
+                Task newTask = new Deadline(taskDescription, deadlineDate);
+                taskList.add(newTask);
+                System.out.println(" Got it. I've added this task:\n"
+                        + "  " + newTask + "\n"
+                        + " Now you have " + taskList.size() + " tasks in the list.");
+            }
+
+            else if (input.startsWith("todo ")) {
```
tanjiajiajun: Having spaces at the end of the string does not look very clean. Maybe can consider removing that.

src/main/java/Bob.java
```diff
@@ -0,0 +1,221 @@
...
+        this.by = by;
+    }
+
+    /*
+     * A method that returns the task status as a string.
+     * @return A label [D] and a check-box followed by the description of the task.
+     */
+    @Override
+    public String toString() {
+        return "[D]" + super.toString() + " (by: " + by + ")";
```
tanjiajiajun: Clean way to handle output. I did not know that!

tanjiajiajun (commented):
Looks great to merge. Just small nits

data/bob.txt
jovantanyk, line 1: Local data files like this should not be pushed onto remote

src/main/java/Bob.java
jovantanyk, line 1: Maybe consider separating your code into different classes and packages :)

src/main/java/Bob.java
```diff
@@ -0,0 +1,430 @@
...
+        catch (IOException e) {
+            System.out.println(e.getMessage());
+        }
+    }
+}
+
+/*
+ * This class represents a task we want to record.
+ */
+class Task {
```
jovantanyk: Different Classes should not be in the same .java file. Consider separating them into dedicated java files

src/main/java/Bob.java
```diff
@@ -0,0 +1,430 @@
...
+    public void markAsUndone() {
+        this.isDone = false;
+    }
+
+    /*
+     * A method that returns the isDone boolean.
+     *
+     * @return A boolean depending on whether the task is done.
+     */
+    public boolean getIsDone() {
```
jovantanyk: By java convention, methods that start with get() usually return a Value like a string, integer or array. For booleans you can just use isDone()

src/main/java/Bob.java
```diff
@@ -0,0 +1,430 @@
...
+/*
+ * This class represents a task we want to record.
+ */
+class Task {
+    private String description;
+    private boolean isDone;
+
+    /*
+     * A constructor that depicts a new task.
+     */
```
jovantanyk: Since Task is never initiated and instead inherited by other subclasses, consider using an abstract class instead.

src/main/java/Bob.java
```diff
@@ -0,0 +1,430 @@
...
+
+                bufferedWriter.write(line);
+                bufferedWriter.newLine();
+            }
+            bufferedWriter.close();
+            System.out.println("File saved.");
+        }
+
+        catch (IOException e) {
+            System.out.println(e.getMessage());
```
jovantanyk: Very basic way of handling exceptions. Consider creating a new Exception class so you can consolidate your exceptions and have different behavior for different types of Exceptions. This method of doing things just means that when your program really does have an exception, it'll be hard to pinpoint where or how it came from because the try block is so big

jovantanyk (commented):
Good attempt however please try and separate out your classes, they should not all be in the same java file. There's also almost no use of OOP here, you can create utility classes for handling Ui, Exceptions, Files, instead of storing them all inside methods.

PR #508, opened 2024-02-12
tcjazwei iP

tcjazwei:


src/main/java/Task.java
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * Constructor
+     * 
+     * @param name task name
+     * @return new task instance
+     */
+    public Task(String name) {
+        this.taskName = name;
+        this.isDone = false;
```
gowthaman-01, line 13: Try to reduce the use of this. Use of this is discouraged, potentially to avoid unnecessary noise in the code. Try to avoid using this except when it is shadowed by a method or constructor parameter.

src/main/java/Dackel.java
```diff
@@ -0,0 +1,382 @@
...
+        return true;
+    }
+
+    public static void main(String[] args) {
+        // title cards, etc.
+        System.out.println(LINE);
+        System.out.println(LOGO);
+        System.out.println(VERSION);
+        System.out.println(LINE);
+        System.out.println();
```
gowthaman-01, line 360: You could extract these print statements into a separate method, to follow the Single Layer of Abstraction Principle (SLAP)

src/main/java/Dackel.java
```diff
@@ -0,0 +1,382 @@
...
+    private static final String UNTIL = "to";
+
+    /** Other String constants */
+    private static final String NAME = "DACKEL";
+    private static final String LOGO = " ____    __    ___  _  _  ____  __   \n"
+        + "(  _ \\  /__\\  / __)( )/ )( ___)(  )  \n"
+        + " )(_) )/(__)\\( (__  )  (  )__)  )(__ \n"
+        + "(____/(__)(__)\\___)(_)\\_)(____)(____) ";
+    private static final String LINE = "____________________________________________________________";
+    private static final String VERSION = "Version 0.1";
```
Arixeyeion, line 30: I like that the programme is personalized to output DACKEL.

src/main/java/Dackel.java
```diff
@@ -0,0 +1,382 @@
...
+
+    /**
+     * Marks the specified task as not done
+     * 
+     * @param index index of task in storedTasks
+     */
+    private static void unmarkTask(int index) {
+        storedTasks.get(index).unmark();
+        speak("i've unmarked the following task. do it soon, please!\n " + storedTasks.get(index).toString());
+    }
```
Arixeyeion, line 150: Should this be extracted out?

I noticed Dackel is used to handle everything from creating tasks to handling errors but I do think it will be easier to debug if you can extract the functionalities into different classes. I suggest having a parser class to parse strings to split up the task creation instead of letting Dackel handle everything.

src/main/java/Event.java
```diff
@@ -0,0 +1,25 @@
...
+        this.startTime = startTime;
+        this.endTime = endTime;
+    }
+
+    @Override
+    public String toString() {
+        String s = "[E]" + super.toString();
+        s = s + " (from " + this.startTime + " ";
+        return s + "until " + this.endTime + ")";
+    }
```
Arixeyeion, line 24: I suggest using String.format to have an easier time producing the string as you can type it out as a sentence that is close to the output.

Arixeyeion (changes requested):
Overall, I like the personalization of DACKEL. I noted some features that could be extracted out so that Dackle does not have to handle everything, this should make the code easier to debug when there are expansions. Also, using String,format can help in making the code more readable and easier to format.

src/main/java/Dackel.java
reetmitra, line 150: I think splitting up the functionality of this class into separate classes would be better? That way, implementing new functionality would be easier and it is also easier to debug if there are errors.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,21 @@
...
+        super(name);
+        this.dueTime = dueTime;
+    }
+
+    @Override
+    public String toString() {
+        String s = "[D]" + super.toString();
+        return s + " (due by " + this.dueTime + ")";
+    }
+}
```
reetmitra, line 21: For the deadline input, consider adding in the functionality to take in date time inputs.

src/main/java/Dackel.java
reetmitra, line 360: For the print statements, create a separate method.

reetmitra (commented):
Project structure is not the greatest, could be split into separate classes for better readability and understanding. Overall, not much coding violations and JavaDocs are written quite well, explaining each method thoroughly.

PR #525, opened 2024-02-12
[Yu Zizhen] ip

YuZizhen:
This is my first pull request to the ip.duke.
will have more in the future.


src/main/java/task/TaskList.java
```diff
@@ -0,0 +1,120 @@
...
+        int length = this.todoList.size();
+        if (length == 1) {
+            System.out.println("Now you have " + length + " task in the list.");
+        } else {
+            System.out.println("Now you have " + length + " tasks in the list.");
+        }
+        storage.autoUpdate(this.todoList);
+    }
+
+    public void findTask(String userInput, Storage storage) {
```
gowthaman-01, line 109: Add JavaDoc to this method

src/main/java/task/TaskList.java
```diff
@@ -0,0 +1,120 @@
...
+                String pos = String.valueOf(i + 1);
+                System.out.println(pos + "." + result.todoList.get(i));
+            }
+        }
+    }
+
+    /**
+     * List the whole TodoList after adding any tasks.
+     * 
+     * @param t
```
gowthaman-01, line 51: Try to add a description of the param as well

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,123 @@
...
+
+    private Task createTaskFromLine(String line) {
+        String[] parts = line.split(" \\| ");
+        if (parts.length >= 3) {
+            boolean isDone = Boolean.parseBoolean(parts[1]);
+            String description = parts[2];
+            DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd");
+
+            switch (parts[0]) {
+                case "T":
```
gowthaman-01, line 99: Try to use an enum instead of string matching

src/main/java/storage/Storage.java
```diff
@@ -0,0 +1,123 @@
...
+    public void autoUpdate(ArrayList<Task> todoList) {
+        File file = new File(this.pathName);
+        file.getParentFile().mkdirs();
+        try {
+            if (!file.exists()) {
+                file.createNewFile();
+            }
+            BufferedWriter writer = new BufferedWriter(new FileWriter(file));
+            for (Task t : todoList) {
+                if (t instanceof Todo) {
```
gowthaman-01, line 51: Avoid using instanceof, instead try to use polymorphism to achieve the same functionality

src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,165 @@
...
+        scanner.close();
+    }
+
+    /**
+     * Handle the three cases when the user input is "todo", "deadline" or "event".
+     * 
+     * @param userInput the input from the user.
+     */
+
+    public void echo(String userInput) {
```
gowthaman-01, line 58: This method is too long, try refactoring it by extracting out some parts into a new method

src/test/java/parser/ParserTest.java
```diff
@@ -0,0 +1,23 @@
...
+
+class ParserTest {
+    @Test
+    public void isMarkTaskTest() {
+        Storage storage = new Storage("./data/duke.txt");
+        ArrayList<Task> temp = new ArrayList<>();
+        temp = storage.getHistory();
+        TaskList TodoList = new TaskList(temp);
+        Parser parser = new Parser(TodoList, storage);
+        assertEquals(parser.isMarkTask("mark 1.2"), false);
```
AwJoey, line 20: Nice test case! Perhaps you could also check if the task can be unmarked?

src/main/java/parser/Parser.java
```diff
@@ -0,0 +1,165 @@
...
+        }
+    }
+
+    /**
+     * Check whether the user is intend to mark one of the tasks.
+     * 
+     * @param userInput
+     * @return True means the user want to mark, or false, which is otherwise.
+     */
+
```
AwJoey, line 118: Great way of organizing your code! 💯  Perhaps you could implement class/enum for the command formatting?

src/main/java/task/TaskList.java
```diff
@@ -0,0 +1,120 @@
...
+        int number = Integer.parseInt(words[1]);
+        Task t = this.todoList.remove(number - 1);
+        System.out.println("Noted. I've removed this task:");
+        System.out.println(t);
+        int length = this.todoList.size();
+        if (length == 1) {
+            System.out.println("Now you have " + length + " task in the list.");
+        } else {
+            System.out.println("Now you have " + length + " tasks in the list.");
+        }
```
AwJoey, line 105: Perhaps you could move this to Ui.java? So that all in/out can be handled there :)

src/main/java/task/Deadline.java
```diff
@@ -0,0 +1,46 @@
...
+
+    private String formatter(LocalDate localDate) {
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("MMM dd yyyy");
+        String formattedDate = localDate.format(formatter);
+        return formattedDate;
+    }
+
+    @Override
+    public String toString() {
+        return "[D]" + super.toString() + " (by: " + formatter(by) + ")";
```
AwJoey, line 44: Maybe you could create a static variable for the tasktype abbreviations (i.e. "D")? This might make it easier to edit in the future if this abbreviation changes

AwJoey (commented):
Great work so far! 💯 Suggested some small tweaks :)

Happy Chinese new year!!

src/main/java/task/TaskList.java
YuZizhen: Ok! I will add on to it. Thanks!

src/main/java/task/TaskList.java
YuZizhen: Sure

src/main/java/storage/Storage.java
YuZizhen: Noted with thank!

src/main/java/storage/Storage.java
YuZizhen: Could you explain a bit more about how to use polymorphism to achieve it? Thanks

src/test/java/parser/ParserTest.java
YuZizhen: Sure! I will add on to the test case

src/main/java/parser/Parser.java
YuZizhen: Oki! Thanks!

src/main/java/task/TaskList.java
YuZizhen: Ah I see! Thank you for pointing out!

src/main/java/task/Deadline.java
YuZizhen: Nice suggestion! Thanks

PR #527, opened 2024-02-13
[ShanylOng] iP

ShanylOng:


src/main/java/Barry.java
```diff
@@ -0,0 +1,322 @@
...
+        }
+
+        // Returns the status of the task
+        protected final boolean getStatus() {
+            return this.isDone;
+        }
+
+        // Returns details about the relevant timings of the task, to be defined in the subclasses
+        abstract public String detailsAsString();
+    }
```
Ty-stan0417, line 36: Consider creating a new file called Task.java

src/main/java/Barry.java
```diff
@@ -0,0 +1,322 @@
...
+
+        // Returns the name of the task
+        protected final String getName() {
+            return this.taskName;
+        }
+
+        // Returns the status of the task
+        protected final boolean getStatus() {
+            return this.isDone;
+        }
```
Ty-stan0417, line 32: Consider getStatus() return a String not a boolean as you are using the getStatus() to determine whether the String should look like.

src/main/java/Barry.java
```diff
@@ -0,0 +1,322 @@
...
+    private static class ToDo extends Task {
+        // Constructor for ToDos, initialised with the name of the task
+        public ToDo(String name) {
+            super(name);
+        }
+
+        // Method which returns details of the object as a String
+        @Override
+        public final String detailsAsString() {
+            return "[T][" + (this.getStatus() ? "X" : " ") + "] " + this.getName();
```
Ty-stan0417, line 49: Consider this.getStatus() to return a String such that if True ? "X" : " ".

src/main/java/Barry.java
```diff
@@ -0,0 +1,322 @@
+import java.util.Scanner;
+import java.util.ArrayList;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class Barry {
+
+    // A class for elements in the taskList, with a name and an indicator of whether the task is marked
+    private abstract static class Task{
```
tahsinhasem, line 9: Consider having a separate file for the `Task` class. I think making the Task class abstract is smart. However, I would like to suggest you make it public/protected - as future specifications may involve using the task class to implement other functionality.

src/main/java/Barry.java
```diff
@@ -0,0 +1,322 @@
...
+        System.out.println("\t   bye                                            Quit");
+        System.out.println("\t   list                                           List all tasks in your list");
+        System.out.println("\t   mark [x]                                       Mark task number x in your list");
+        System.out.println("\t   unmark [x]                                     Unmark task number x in your list");
+        System.out.println("\t   delete [x]                                     Delete task number x in your list");
+        System.out.println("\t   todo [task name]                               Add a ToDo to your list");
+        System.out.println("\t   deadline [task name] /by [deadline]            Add a Deadline to your list");
+        System.out.println("\t   event [task name] /from [start] /to [end]      Add an Event to your list");
+        System.out.println("\t____________________________________________________________\n");
+    }
```
Ty-stan0417, line 125: LGTM. I like the way you have the help() function to list all available commands to users. Good job!

tahsinhasem (commented):
Nice work! However, some suggestions regarding the coding standards and my opinions -

src/main/java/Barry.java
```diff
@@ -0,0 +1,322 @@
...
+            this.deadlineDateTime = deadline;
+        }
+
+        // Method which returns details of the object as a String
+        @Override
+        public final String detailsAsString() {
+            String basicInfo = "[D][" + (this.getStatus() ? "X" : " ") + "] " + this.getName();
+            return basicInfo + " (by: " + deadlineDateTime + ")";
+        }
+    }
```
Ty-stan0417, line 70: Consider to create a new file called Deadline.java and move the code to the file.

src/main/java/Barry.java
```diff
@@ -0,0 +1,322 @@
...
+            this.eventEndDateTime = end;
+        }
+
+        // Method which returns details of the object as a String
+        @Override
+        public final String detailsAsString() {
+            String basicInfo = "[E][" + (this.getStatus() ? "X" : " ") + "] " + this.getName();
+            return basicInfo + " (from: " + eventStartDateTime + " to: " + eventEndDateTime + ")";
+        }
+    }
```
Ty-stan0417, line 91: Consider to create a new file called Event.java and move the code there.

src/main/java/Barry.java
```diff
@@ -0,0 +1,322 @@
...
+                    System.out.println("\t Looks like there aren't any tasks in your list yet... let's add some!");
+                } else {
+                    // List has items
+                    System.out.println("\t Let's take a look at the tasks in your list:");
+                    // Print list of tasks
+                    for (int i = 0; i < taskListSize; i++) {
+                        Task task = taskList.get(i);
+                        System.out.println("\t " + (i + 1) + "." + task.detailsAsString());
+                    }
+                }
```
Ty-stan0417, line 162: Consider to create a new method that can print out the list in your class so that you can just use it here.

src/main/java/Barry.java
```diff
@@ -0,0 +1,322 @@
...
+            this.isDone = false;
+        }
+
+        // Allows the user to modify the status of the task
+        public void modifyStatus(boolean status) {
+            this.isDone = status;
+        }
+
+        // Returns the name of the task
+        protected final String getName() {
```
tahsinhasem, line 25: Consider making the functions getName and modifyStatus have the same privacy - either both protected or both public.

src/main/java/Barry.java
```diff
@@ -0,0 +1,322 @@
...
+        System.out.println("\t   delete [x]                                     Delete task number x in your list");
+        System.out.println("\t   todo [task name]                               Add a ToDo to your list");
+        System.out.println("\t   deadline [task name] /by [deadline]            Add a Deadline to your list");
+        System.out.println("\t   event [task name] /from [start] /to [end]      Add an Event to your list");
+        System.out.println("\t____________________________________________________________\n");
+    }
+
+    // Task list mode, allows user to enter tasks into a list, and print all tasks in the list
+    // currently with "list", exits with "bye"
+    private static void taskListMode() {
```
Ty-stan0417, line 129: LGTM now. But later in the future, you may be facing on how to separate the code in this taskListMode() method. You may try to separate the execution of each of the command to its class. For an example, you can only mark or unmark a task, you may consider adding methods in your Task class, to unmark or mark the specific task. Then, the code in this method when the command is mark or unmark you can just simply using the method you created and making the code shorter. This will be allow you easy to change while you are in increments A-More-OOP later.

src/main/java/Barry.java
```diff
@@ -0,0 +1,322 @@
...
+            return this.taskName;
+        }
+
+        // Returns the status of the task
+        protected final boolean getStatus() {
+            return this.isDone;
+        }
+
+        // Returns details about the relevant timings of the task, to be defined in the subclasses
+        abstract public String detailsAsString();
```
tahsinhasem, line 35: Nice work with this!

src/main/java/Barry.java
tahsinhasem, line 9:
```suggestion
    private abstract static class Task {
```

src/main/java/Barry.java
```diff
@@ -0,0 +1,322 @@
...
+                System.out.println("\t____________________________________________________________\n");
+            }
+        }
+    }
+
+    public static void main(String[] args) {
+        greet();
+        taskListMode();
+        exit();
+    }
```
Ty-stan0417, line 321: LGTM. I like the way you have code the main method will fewer lines. Consider using polymorphism, overloading and overriding when you coding in other method. Overall good job!

src/main/java/Barry.java
```diff
@@ -0,0 +1,322 @@
...
+    // Task list mode, allows user to enter tasks into a list, and print all tasks in the list
+    // currently with "list", exits with "bye"
+    private static void taskListMode() {
+        Scanner scanner = new Scanner(System.in);
+        final Pattern markPattern = Pattern.compile("^mark\\s(\\d+)$");
+        final Pattern unmarkPattern = Pattern.compile("^unmark\\s(\\d+)$");
+        final Pattern deletePattern = Pattern.compile("^delete\\s(\\d+)$");
+        final Pattern todoPattern = Pattern.compile("^todo\\s(.+)$");
+        final Pattern deadlinePattern = Pattern.compile("^deadline\\s(.+)\\s/by\\s(.+)$");
+        final Pattern eventPattern = Pattern.compile("^event\\s(.+)\\s/from\\s(.+)\\s/to\\s(.+)$");
```
Ty-stan0417, line 136: LGTM. I like the way you listed down all the command pattern. Good!

src/main/java/Barry.java
```diff
@@ -0,0 +1,322 @@
...
+
+    // Exits the program with a message
+    private static void exit() {
+        System.out.println("\t____________________________________________________________");
+        System.out.println("\t Bye, see you next time! :)");
+        System.out.println("\t____________________________________________________________\n");
+    }
+
+    // List all available commands
+    private static void help() {
```
tahsinhasem, line 112: Like the help command! Nice! However, maybe instead of manually adding space, consider adding `\t`

src/main/java/Barry.java
```diff
@@ -0,0 +1,322 @@
...
+        // Execution loop
+        while (true) {
+            // Command line input, stripped of whitespaces
+            String inputText = scanner.nextLine().strip();
+            if (inputText.equals("bye")) {
+                // Exit task list mode
+                break;
+            } else if (inputText.equals("help")) {
+                // List all available commands
+                help();
```
Ty-stan0417, line 146: LGTM. I like the way you doing here. Using the help() method from class to doing listed down all the commands. Consider the others commands to make your code shorter here.

Ty-stan0417 (commented):
Looks good. 👌
But you need to catch up your iP progress.
Remember when you create a new class, try creating a new file also.
Don't forget to use overriding, overloading and polymorphism. These will make your code looks more better.

src/main/java/Barry.java
```diff
@@ -0,0 +1,322 @@
...
+                    try {
+                        Task task = taskList.get(taskIndex);
+                        if (!task.getStatus()) {
+                            task.modifyStatus(true);
+                            System.out.println("\t Nice, I have marked this task as done:");
+                            System.out.println("\t   " + task.detailsAsString());
+                            System.out.println("\t Great job! :)");
+                        } else {
+                            System.out.println("\t Looks like this task was already marked...");
+                            System.out.println("\t [In order to unmark a task, use the 'unmark' command instead]");
```
tahsinhasem, line 180: That is helpful!

src/main/java/Barry.java
```diff
@@ -0,0 +1,322 @@
...
+                        System.out.println("\t Oops, I can't find a task in your list with that number! :(");
+                        System.out.println("\t [Make sure the task number is correct and try again]");
+                    }
+                } else {
+                    // Invalid format
+                    System.out.println("\t I see you're trying to mark a task, but it's in the wrong format...");
+                    System.out.println("\t [Type 'help' if you're unsure of the correct format]");
+                }
+                System.out.println("\t____________________________________________________________\n");
+            } else if (inputText.startsWith("unmark")){
```
tahsinhasem, line 193:
```suggestion
            } else if (inputText.startsWith("unmark")) {
```

src/main/java/Barry.java
```diff
@@ -0,0 +1,322 @@
...
+                    // List has items
+                    System.out.println("\t Let's take a look at the tasks in your list:");
+                    // Print list of tasks
+                    for (int i = 0; i < taskListSize; i++) {
+                        Task task = taskList.get(i);
+                        System.out.println("\t " + (i + 1) + "." + task.detailsAsString());
+                    }
+                }
+                System.out.println("\t____________________________________________________________\n");
+            } else if (inputText.startsWith("mark")) {
```
tahsinhasem, line 164: Cosider Having functions inside the else if block which deals with the logic instead of having a big block of logic. That will make the code more readable. EG:

```suggestion
            } else if (inputText.startsWith("mark")) {
                handleMark(inputText);
            }
```

src/main/java/Barry.java
```diff
@@ -0,0 +1,322 @@
+import java.util.Scanner;
+import java.util.ArrayList;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class Barry {
+
+    // A class for elements in the taskList, with a name and an indicator of whether the task is marked
```
jasperng-nus, line 8: I think perhaps you have no reached the part of the iP where they ask you to add JavaDocs, but I like that you left a comment on the class to summarise what it does, could implement JavaDocs once you reach the stage

tahsinhasem (commented):
Neat work! Follows the coding standards well. However, some overall comments, apart from ones provided:
- Please consider breaking the program into classes and packages. It will improve code readability.
- Nice work describing the methods with comments. Maybe you can add javadoc support.

src/main/java/Barry.java
jasperng-nus, line 36: I like the idea of making the Task class abstract however, I think you could create another file for the Task class and perhaps package the classes that extends from Task into a folder.

src/main/java/Barry.java
jasperng-nus, line 70: Can consider creating a separate file for Deadline

src/main/java/Barry.java
jasperng-nus, line 91: Can considering create another file for Event

src/main/java/Barry.java
jasperng-nus, line 32: I think you could consider implementing your conditional statement in the getStatus() instead of returning boolean, as in the detailsAsString() method, you did the conditional statement in the additions of strings. I think it could perhaps be something like getStatusString() and you will get the string directly instead of having conditional statements in the additions of strings.

src/main/java/Barry.java
jasperng-nus, line 125: I like that you put the UIs into different methods but perhaps I think you could create a Ui class which encapsulates all this methods together. Also I like that you have a help menu it will really help the user out!

src/main/java/Barry.java
jasperng-nus, line 136: LGTM. I like the idea of this and how it uses regex to capture the different patterns.

src/main/java/Barry.java
jasperng-nus, line 162: I think you could create a method similar to your Ui commands except that it takes in the current list and it you could put these codes into the new method.

src/main/java/Barry.java
```diff
@@ -0,0 +1,322 @@
...
+                System.out.println("\t I'm sorry, I don't understand what that means... >_<");
+                System.out.println("\t [Type 'help' for a list of available commands]");
+                System.out.println("\t____________________________________________________________\n");
+            }
+        }
+    }
+
+    public static void main(String[] args) {
+        greet();
+        taskListMode();
```
jasperng-nus, line 319: I like that you main method does not have much lines, but I think you could consider creating other class, using packages to structure your files better as well, especially in the taskListMode() method where it looks a bit clunky, I think you consider create perhaps a Command package and add the different command classes. Overall, I like the idea here and, keep up the good work!

jasperng-nus (commented):
Overall, good job on creating a nice user interface and compacting it all into methods! I have added some comments you can take look and see if it helps!

PR #532, opened 2024-02-13
[Ryan-Reno] iP

Ryan-Reno:
LITE

Lite is the next generation's ToDo List! It's :
- text-based
- ~~EASY~~ SUPER EASY to learn and use

All you need to do is,
1. download it from [here](https://github.com/Ryan-Reno/ip).
2. double-click it.
3. add your tasks.
4. let it keep track of all the tasks for you 😃 

And it is **COMPLETELY FREE!**

Features :
- [x] Managing tasks
- [x] Managing deadlines
- [ ] Reminders (coming soon) 

```
        Chatbot chatbot = new Chatbot();
        chatbot.start();
 }
```


src/main/java/Chatbot.java
```diff
@@ -0,0 +1,208 @@
...
+            return;
+        }
+
+        if (instruction[0].equals("unmark")){
+            this.unmark(instruction);
+            return;
+        }
+
+        if (instruction[0].equals("delete")) {
+            this.delete(instruction);
```
AjayArvind2207: Consider removing the 'this' reference since there is inconsistent styling with addToDo among other instructions
```suggestion
            delete(instruction);
```

src/main/java/Chatbot.java
```diff
@@ -0,0 +1,208 @@
+import java.sql.Array;
+import java.util.Scanner;
+import java.util.ArrayList;
+
+public class Chatbot {
+    private Scanner scanner = new Scanner(System.in);
+    private String currentInput;
+    private ArrayList<Task> list;
```
AjayArvind2207: Suggestion: Why not have a Tasks class to encapsulate the ArrayList<Task> objects? That way it makes the code in Chatbot.java a lot less dense, and you can abstract away most of the complexity into that class.

src/main/java/Lite.java
```diff
@@ -0,0 +1,15 @@
+public class Lite {
+    public static void main(String[] args) {
+        String logo = "  LLLLL      IIIIIIIII  TTTTTTTTTT  EEEEEEEEEE  \n" +
```
AjayArvind2207: This is super cool and I really enjoy it :D

AjayArvind2207 (commented):
Code looks good and well written. I did feel as though you could have abstracted a bit more in terms of classes -- using smaller classes to encapsulate behaviour you have used in Chatbot.java.

src/main/java/Chatbot.java
Ryan-Reno: Sorry I don't quite get what you mean. Isn't the point of the Task class to encapsulate the ArrayList objects? If so, doesn't my Task class behave the same with the Tasks class you suggested?

src/main/java/lite/Ui.java
jovantanyk, line 1: Method Names could be clearer. parse() could be renamed into executeCommand() instead. 
Methods like find, delete, mark addDeadline, addEvent could be renamed into findTask, deleteTask, markTaskAsDone, addDeadlineTask, addEventTask for more clarity.
Overall not a bad attempt however more OOP could be used. The Ui class is currently parsing input strings, processing business logic, and handling output all at the same time. You could separate the command parsing into a different class and allow the Ui class to solely handle text generation

src/main/java/lite/util/Printer.java
jovantanyk, line 1: Method names seem inconsistent. toDoException could be renamed to printToDoException. successfulFind and unsuccessfulFind could be renamed to printFound() and printNotFound(). Since the Printer class solely handles output, methods should have consistent naming.
Also, it seems redundant to have methods to print Exceptions here when you're also handling exceptions in LiteExceptions.

src/main/java/lite/task/TaskList.java
jovantanyk, line 1: Method names could be better. remove() could be more specific like removeTask(). taskString() is fine but toString() would be better.

jovantanyk (commented):
Overall Good attempt but method names could use more consistent naming. UI class is currently handling too many functions and could be split up more. Exception handling is a bit weird as it doesn't actually catch any exceptions. LiteException class should extend Exception so you can actually use try catch blocks to handle exceptions.

src/main/java/lite/Ui.java
```diff
@@ -43,12 +45,33 @@ public boolean parse(TaskList tasks) {
...
             this.addEvent(instruction, tasks);
+        } else if (instruction[0].equals("find"))  {
+            this.find(instruction, tasks);
         } else {
             LiteException.invalidInput();
         }
         return false;
     }
 
+    private void find (String instruction[], TaskList tasks) {
```
jovantanyk: method name could be clearer for example findTask() instead of just find()

src/main/java/lite/task/TaskList.java
```diff
@@ -59,6 +68,15 @@ public void outputTasks() {
         Printer.printHorizontalLine();
     }
 
+    /**
+     * Outputs the string representation of the TaskList
+     */
+    public void taskString() {
```
jovantanyk: toString() would be better and follow naming conventions

src/main/java/lite/util/Printer.java
Ryan-Reno: Thanks for this feedback! Was not thinking properly when creating the methods and simply put it within the Printer class.

PR #560, opened 2024-02-16
B-Reminders Extension (BCD-Extension)

SherisseTJW:
Features:
- Outputs reminders of upcoming due tasks to the user upon startup
- Introduces a new 'reminder' command that will also output all reminders of upcoming due tasks


SherisseTJW:
Oops, sorry

